PR #4, opened 2019-11-06
WIP: Add type for qr transaction data

phieronymus:
- Add qr transaction brand type
- Add qr transaction gateway type
- Add types to transaction data


phieronymus:
Created a fork as I do not have the permission to add a branch on the repository directly. It should not after the merge.

phieronymus:
Do not merge yet. I should double check the values.

PR #9, opened 2020-02-03
Add cancel condition to long polling

phieronymus:
Updates https://github.com/univapaycast/univapay-console/issues/1883

### DIFF
--- a/src/api/RestAPI.ts
+++ b/src/api/RestAPI.ts
@@ -307,10 +307,11 @@ export class RestAPI {
     async longPolling<A>(
         promise: PromiseCreator<A>,
         condition: (response: A) => boolean,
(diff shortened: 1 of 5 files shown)


src/api/RestAPI.ts
```diff
@@ -325,8 +326,12 @@ export class RestAPI {
                 (async function repeater(): Promise<A> {
                     const result = await promise();
 
+                    if (cancelCondition && cancelCondition(result)) {
```
laxa88: hmm, just wondering, is it possible to write `cancelCondition?.(result)` in the conditional here?

phieronymus: We are not using typescript 3.7 on this project and I was afraid to break something by doing the update 🤔

PR #10, opened 2020-03-11
add update param for bank account

laxa88:
this prop is missing in the definition


phieronymus:
Merging this change! Thanks 🙂

PR #19, opened 2020-06-17
update Configuration prop

laxa88:
Related to https://github.com/univapaycast/univapay-admin-console/issues/284


src/resources/common/types.ts
```diff
@@ -18,6 +18,7 @@ export interface WithConfig<T> {
 export interface AmountWithCurrency {
     amount: number;
     currency: string;
+    duration?: string;
```
laxa88, line 21: this shouldn't be here but the API returns this value, and we're using it in our code, so... 🤷‍♂️

PR #27, opened 2020-07-17
Add IssuerUrl resource

perrin4869:
Missing tests yet


phieronymus:
Looking good. However you may want this route to be in the same file as the other charges route as the resource belongs to them 🙂

phieronymus:
Adding me and @laxa88 to the list of reviewers for this Pull Request 🙂

perrin4869:
OK, moved to the `charges` route. BTW, not sure about the response format. Tried to guess based on Kingston's hints!

perrin4869:
Good question! No idea lol
In Slack @kgston said that the response format is

```scala
implicit val format: OFormat[OnlineData] = (
    (__ \ "issuer_token").formatOption[String] ~
      (__ \ "http_method").formatOption[String]
  )(OnlineData.apply, unlift(OnlineData.unapply))
```

Which I just assumed means it gets an `httpMethod` property, but maybe I just can't read this format thing.
BTW, I also marked the returned properties `storeId` and `chargeId`, but I notice now that maybe they aren't being set afterall

kgston:
The `http_method` tells you which http verb to use when making the request to the `issuer_token` which contains the URL you need to call. It can be `GET`, but sometimes it can be `PUT` or `POST`.

perrin4869:
Went ahead and added `TransactionTokenOnlineData` inside this PR too

phieronymus:
Merging this. Thanks for the update!

PR #37, opened 2020-07-28
Add EventEmitter to SDK

### LINKED ISSUE #35: Add either logger or events to the SDK object
Some SDKs like `alipay` provide a `logger` property that you can supply, such as `console` or a [winston](https://github.com/winstonjs/winston) logger, etc, so that before each request/response, the SDK logs out. It would be a useful feature to add here as well.

```javascript
import logger from "services-logger";
import getAlipayClient from "./alipay";

const alipay = getAlipayClient(
    {
        appId: ALIPAY_APP_ID,
        privateKey: ALIPAY_PRIVATE_KEY,
        alipayPublicKey: ALIPAY_PUBLIC_KEY,
        gateway: ALIPAY_GATEWAY,
        keyType: "PKCS8"
    },
    { logger: logger("alipay") } // can also specify { logger: console }, or any object with a debug, info, etc function
);
```

There are 4 different approaches I can think of here:

1. Do the same as alipay and provide a `logger` option
2. A more flexible approach is to extend `EventEmitter` and emit `request` and `response` events with the appropriate data, and the user can use these to log out as they see fit
3. Add properties such as `{ onRequest: (string) => void, onResponse: string => void }` which are basically event emitters as well
4. Discard this feature and leave this up to userland

Any opinions?

perrin4869:
Closes #35 

Uses [eventemitter3](https://github.com/primus/eventemitter3), and is completely type-safe
Super straightforward :D

### DIFF
--- a/test/specs/api.specs.ts
+++ b/test/specs/api.specs.ts
@@ -202,7 +202,6 @@ describe("API", function () {
         await api.send(HTTPMethod.GET, "/ok");
 
         expect(onResponse).to.have.been.calledOnce;
-        console.log(onResponse.firstCall.lastArg);
         expect(onResponse.firstCall.lastArg)
             .to.include({
                 status: 200,
--- a/test/specs/api.specs.ts
+++ b/test/specs/api.specs.ts
@@ -166,6 +166,52 @@ describe("API", function () {
         }
     });
 
+    it("should emit request event", async function () {
+        fetchMock.getOnce(`${testEndpoint}/ok`, {
+            status: 200,
+            body: okResponse,
+            headers: { "Content-Type": "application/json" },
+        });
+
+        const onRequest = sinon.spy();
+
+        const api: RestAPI = new RestAPI({ endpoint: testEndpoint });
+        api.on("request", onRequest);
+        await api.send(HTTPMethod.GET, "/ok");
+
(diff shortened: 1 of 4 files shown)


test/specs/api.specs.ts
```diff
@@ -166,6 +166,52 @@ describe("API", function () {
...
+        });
+
+        const onResponse = sinon.spy();
+
+        const api: RestAPI = new RestAPI({ endpoint: testEndpoint });
+        api.on("response", onResponse);
+        await api.send(HTTPMethod.GET, "/ok");
+
+        expect(onResponse).to.have.been.calledOnce;
+        console.log(onResponse.firstCall.lastArg);
```
laxa88, line 205: looks like unnecessary `console.log` ?

perrin4869: oh shit lol thanks!!

perrin4869: a bit too trigger happy :P

phieronymus:
Let me merge this and keep this Pull Request somewhere in case we need to know what events we have 🙂 However the types should be sufficient.

perrin4869:
Yeah the types are really good in this case!!! I think it enforces the types of the handler parameters too :)

PR #39, opened 2020-07-29
Support event emitters from PaymentsSDK and Resource

### LINKED ISSUE #38: Allow event listeners to be used on the SDK
Make `univapay.charges.api.on("request", (request: Request) => logger?.debug(request));`, `univapay.transactionTokens.api.on("request", (request: Request) => logger?.debug(request));`, be possible to do via `univapay.on("request", (request: Request) => logger?.debug(request));` and `univapay.charges.on("request", (request: Request) => logger?.debug(request));`. Maybe via proxy?

perrin4869:
Closes #38 

This works by using the [`newListener`](https://nodejs.org/api/events.html#events_event_newlistener) and [`removeListener`](https://nodejs.org/api/events.html#events_event_removelistener) events - whenever the `Resource` or `PaymentsSDK` gets an event handler, it is relayed to the underlying `api` object.
However, [`eventemitter3` doesn't support these built-in events](https://github.com/primus/eventemitter3/issues/50), so it had to be replaced with a more straightforward [`node` polyfill](https://www.npmjs.com/package/events). The good thing here is that in node environments, we'll just keep using the regular node module. The bad news is that [standard node typings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/v12/events.d.ts) are not as strict as the [`eventemitter3` typings](https://github.com/primus/eventemitter3/blob/master/index.d.ts) - these do not support generics, so the events are not type safe anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,18 @@ const customSdk = new CustomSDK();
 
 ### API Documentation
 
+#### Events
+
+The `PaymentsSDK` and `Resource` based classes such as `Charges` and `Stores` are [EventEmitters](https://nodejs.org/api/events.html). You can subscribe to the following events:
+
+```
+const sdk = new SDK();
+sdk.on('request', (req: Request) => void)
+sdk.on('response', (res: Response) => void)
+```
+
+`Request` and `Response` are the [`fetch` API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) types.
+
 WIP
 
 ### TypeScript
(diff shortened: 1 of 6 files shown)


phieronymus:
Nice trick 🙂 Would it be possible to add the event it can listen to in the `Readme`? As the type disappeared I am scared that finding what the logging is capable of will be harder to find.

perrin4869:
Done!

phieronymus:
Thanks a lot! Merging this 🙂

PR #41, opened 2020-07-30
Override auth parameters on individual api calls

perrin4869:
Blocks https://github.com/univapaycast/services-typescript/issues/25

This is needed in the `miniapp` API to make requests to different `stores` with the same `SDK` object.
The implementation was, as usual, easier than anticipated.
On the `API` surface, this is only breaking if using the `callback` style API calls, since the `callback` argument in all resources moved one step back to make space for `auth` parameters.
Under the surface, I simplified the `pathParameters` mechanism and moved the parameter order slightly, but the tests all pass just the same!
The heart of this PR is in [this](https://github.com/univapay/univapay-node/blob/b9eaaf9c5f4a9b6ee9a34cb5636995cff1b049c7/src/api/RestAPI.ts#L303) line, where I will set the auth params passed directly into the method to the highest priority.

### DIFF
--- a/src/api/RestAPI.ts
+++ b/src/api/RestAPI.ts
@@ -209,6 +209,7 @@ export class RestAPI extends EventEmitter {
         method: HTTPMethod,
         uri: string,
         data?: SendData<Data>,
+        auth?: AuthParams,
         callback?: ResponseCallback<A>,
         requireAuth = true,
         acceptType?: string
@@ -226,7 +227,7 @@ export class RestAPI extends EventEmitter {
         const payload: boolean = [HTTPMethod.GET, HTTPMethod.DELETE].indexOf(method) === -1;
 
         const params: RequestInit = {
-            headers: this.getHeaders(data, payload, acceptType),
+            headers: this.getHeaders(data, auth, payload, acceptType),
             method,
         };
 
@@ -243,7 +244,7 @@ export class RestAPI extends EventEmitter {
 
             this.emit("response", response);
 
-            const jwt = await extractJWT(response);
+            const jwt = extractJWT(response);
 
             if (jwt) {
                 this.jwtRaw = jwt;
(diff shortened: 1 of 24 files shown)


src/api/RestAPI.ts
```diff
@@ -298,6 +300,7 @@ export class RestAPI extends EventEmitter {
 
         // Deprecated
         const { authToken = this.authToken, appId = this.appId, secret = this.secret, jwt = this.jwtRaw } = {
+            ...auth,
             ...(!isFormData ? data : {}),
         };
```
phieronymus, line 305: We may want to remove the `Deprecated` as it is not about form data only any longer.

Also I would recommend to have `...auth` on the second line as it is an override. As formData is not initially used for this purpose it make sense for the `auth` parameter to have the last word in my humble opinion 🙂

src/resources/TemporaryTokenAlias.ts
```diff
@@ -77,28 +77,36 @@ export class TemporaryTokenAlias extends CRUDResource {
...
+        auth?: AuthParams,
         callback?: ResponseCallback<Blob>
     ): Promise<Blob>;
     get(
         storeId: string,
         id: string,
         data?: SendData<TemporaryTokenAliasParams>,
+        auth?: AuthParams,
         callback?: ResponseCallback<ResponseTemporaryTokenAlias>
     ): Promise<ResponseTemporaryTokenAlias>;
```
phieronymus, line 103: Not for this PR. Just curiosity. Is it for typing? 🤔 Looks like a pretty weird structure as the `get` function is implemented after.

perrin4869: Wondered about this one too while refactoring. Honestly don't know, is this a TS override?

src/api/RestAPI.ts
perrin4869: Yup, right! Done!

src/resources/TemporaryTokenAlias.ts
phieronymus: I'll try to remember this weird thing. We may want to make it better at one point (but not in this PR). Just wanted to know if you found out something about his. 🙂

phieronymus:
Merging this. Thanks! 🙂

PR #42, opened 2020-07-31
Improve serialize error resilience

phieronymus:
No issue related to this hotfix.

This solves a lesser issue while using the serializing feature on the admin-console.

Also took to opportunity to replace the redundant Omit by a fix type.

### DIFF
--- a/src/errors/RequestResponseError.ts
+++ b/src/errors/RequestResponseError.ts
@@ -4,15 +4,17 @@
 
 import { ErrorResponse } from "../api/RestAPI";
 
-const serializeErrorResponse = ({ code, httpCode, errors }: Omit<ErrorResponse, "status">): string =>
-    `Code: ${code}, HttpCode: ${httpCode}, Errors: ${errors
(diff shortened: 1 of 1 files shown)


src/errors/RequestResponseError.ts
```diff
@@ -4,30 +4,32 @@
 
 import { ErrorResponse } from "../api/RestAPI";
 
-const serializeErrorResponse = ({ code, httpCode, errors }: Omit<ErrorResponse, "status">): string =>
-    `Code: ${code}, HttpCode: ${httpCode}, Errors: ${errors
+type ErrorRequest = Omit<ErrorResponse, "status">;
+
+const serializeErrorResponse = ({ code, httpCode, errors }: ErrorRequest): string =>
+    `Code: ${code}, HttpCode: ${httpCode}, Errors: ${(errors || [])
```
phieronymus, line 10: This is the only logic change `errors` -> `(errors || [])`. `errors` can be undefined in some cases such as the gateway credentials.

PR #43, opened 2020-07-31
Update TransactionTokens.ts

laxa88:
It seems the order is important... Otherwise TS complains about missing required props

@perrin4869 fyi only. I didn't know about this until I tried to consume it in our console project. 🤔


laxa88:
Hmm... possibly unrelated. I will investigate further before closing this PR.

laxa88:
False alarm. This PR is not necessary.

PR #53, opened 2020-08-25
remove filter logic for SendData, refactor getHeaders

laxa88:
Updates https://github.com/univapaycast/univapay-admin-console/issues/750

### DIFF
--- a/src/api/RestAPI.ts
+++ b/src/api/RestAPI.ts
@@ -84,12 +84,12 @@ export type PromiseReject = (reason?: any) => void;
 export interface AuthParams {
     jwt?: string;
     secret?: string;
+    idempotentKey?: string;
+    origin?: string;
 
     // Deprecated
     authToken?: string;
     appId?: string;
-    idempotentKey?: string;
-    origin?: string;
 }
 
 export interface PollParams {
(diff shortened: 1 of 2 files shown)


src/api/RestAPI.ts
```diff
@@ -88,51 +88,22 @@ export interface AuthParams {
     // Deprecated
     authToken?: string;
     appId?: string;
+    idempotentKey?: string;
```
perrin4869, line 91: Are `idempotentKey` and `origin` deprecated?

perrin4869 (commented):
Just wondering if the new `AuthParam` keys should be placed under the `deprecated` comment!

src/api/RestAPI.ts
laxa88: Not sure, but I didn't see them being used anywhere... But I'll move this away from `deprecated` just in case.

perrin4869: cool!

perrin4869:
want to approve but can't find a button ><;; lol

src/api/RestAPI.ts
```diff
@@ -288,21 +258,23 @@ export class RestAPI extends EventEmitter {
             headers.append("Content-Type", "application/json");
         }
 
-        const origin = getOrigin(data) || this.origin;
+        const {
+            origin = this.origin,
+            idempotentKey,
```
phieronymus, line 263: If you want to follow the old behavior you may want to default assign it to `null` 🙂

phieronymus (approved):
Just a lesser comment in case it was not on purpose 🙂 Otherwise approved!

src/api/RestAPI.ts
laxa88: ok sure, fixed! 🙇

PR #59, opened 2020-09-04
Support ESM and CJS from a single package

perrin4869:
Discussed this briefly about two weeks ago, finally wrapping up the miniapp API got around to doing this.
The main breaking change here is that if you want to import from a file outside the main import, you are going to have to change all import paths.
There are 3 ways I can think of to go about this:

1. Update all import paths (breaking change, this is the path that I took here as we discussed before)
2. Export all relevant functions/types from the main script file. This is how most SDKs/libs are structured. I think I like this better, but it's also a big breaking change.
3. Under node 14 and all packagers that support this, adopt [this](https://2ality.com/2019/10/hybrid-npm-packages.html) approach. Unfortunately, not sure if webpack or rollup support this. I tried this on node 13 before, and it worked like charm, but this requires the bundlers to actually understand this syntax. Would need to research if they do


perrin4869:
Added `exports` to `package.json` because it doesn't hurt to have it there - it can help the bundlers figure things out!
I tested this with `node14` and I could import both `require('univapay-node')` and `require('univapay-node/resources/BankAccounts.js')` after installing this into a bare npm package.

perrin4869:
https://github.com/webpack/webpack/issues/4674#issuecomment-665507118
`webpack@5` seems to support the `exports` field. Rollup doesn't.

laxa88 (approved):
LGTM 🎉

perrin4869:
Oh one thing that I was thinking, will probably be necessary to update the releasing logic, probably in the CI/CD pipeline (jenkins?) and also deprecate `univapay-node-es` afterwards.

phieronymus:
Merging this. I will update the CI logic and handle https://github.com/univapay/univapay-node/issues/64. Thanks 🙂

PR #68, opened 2020-10-01
Feature/pierreh/64/add top level imports

### LINKED ISSUE #64: Add top-level exports
Will make it easier to use the SDK without reaching for deep-level exports which may be tricky to use depending on bundler/platform

phieronymus:
Closes #64 

As I am not used to this, please feel free to tell me if I do something wrong even if that mean re-doing the whole thing 🙂


.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2
 defaults: &defaults
   working_directory: ~/repo
   docker:
-    - image: circleci/node:12
+    - image: circleci/node:14
```
phieronymus, line 6: This is unrelated but I am pushing for all of our modules to be built on node 14 🙂

phieronymus:
This is for the coverage: https://github.com/univapay/univapay-node/issues/69. It was broken at some point 🙂

.circleci/config.yml
```diff
@@ -74,11 +74,7 @@ jobs:
 
       - run:
           name: Publish package LIB
-          command: npm publish lib
-
-      - run:
-          name: Publish package ES
-          command: npm publish es
+          command: npm publish dist
```
perrin4869, line 77: this is just `npm publish`!

perrin4869 (commented):
Previously we were building two separate packages under `lib` and `es`, but now we are publishing a single package at the top level, so it's just `npm publish` (without `dist`)

.circleci/config.yml
phieronymus: Thanks. That makes sense 🙂 

Somehow I forgot to merge pull `master` first and looking at the `package.json` it was requiring the index 🤦 My brain was probably very slow yesterday.

laxa88 (approved):

perrin4869 (approved):

PR #74, opened 2020-10-21
fix unhandled error type in serializeErrorResponse

laxa88:
No associated ticket.

Fixes/improves handling non-object error types returned from API (i.e. `string` types).


phieronymus (approved):

perrin4869:
Hey, just noticed this one, I know it's merged, but I was thinking maybe it's a bit better to test `typeof error === "string"` instead? It's a bit more conservative

PR #80, opened 2020-11-25
Add cvv authorized to transaction token item

phieronymus:
Updates https://github.com/univapaycast/univapay-admin-console/issues/1063

### DIFF
--- a/src/resources/TransactionTokens.ts
+++ b/src/resources/TransactionTokens.ts
@@ -60,6 +60,13 @@ export enum RecurringTokenPrivilege {
     INFINITE = "infinite",
 }
 
+export enum CvvAuthorizedStatus {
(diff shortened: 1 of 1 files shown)


src/resources/TransactionTokens.ts
```diff
@@ -176,8 +194,11 @@ export interface TransactionTokenConvenienceDataItem {
     customerName?: string;
     expirationPeriod?: string;
     phoneNumber?: PhoneNumber;
+    cvvAuthorize?: {
+        enabled: boolean;
+        status: CvvAuthorizedStatus;
+    };
 }
-
```
laxa88, line 180: not sure why this line was removed, but anyway not important 🤔

laxa88 (approved):

PR #81, opened 2020-11-25
update defs for paidy data

laxa88:


src/resources/TransactionTokens.ts
```diff
@@ -150,6 +151,14 @@ export interface TransactionTokenCardDetails {
     category?: CardCategory;
 }
 
+export interface TransactionTokenPaidyBilling {
+    line1?: string;
+    line2?: string;
+    city?: string;
+    state?: string;
+    zip: string;
+}
```
phieronymus, line 160: No country? 🙂

phieronymus (approved):

src/resources/TransactionTokens.ts
laxa88: https://paidy.com/docs/en/paidycheckout.html#buyer_pay

According to the docs, not required

PR #85, opened 2020-11-30
Add public ecform data

phieronymus:
Updates https://github.com/univapaycast/univapay-form-checkout/issues/85

### DIFF
--- a/src/resources/ECForms.ts
+++ b/src/resources/ECForms.ts
@@ -1,0 +1,121 @@
+/**
+ *  @module Resources/ECForms
+ */
+import { AuthParams, ResponseCallback, SendData } from "../api/RestAPI";
+
+import { AmountWithCurrency } from "./common/types";
+import { CRUDResource } from "./CRUDResource";
+import { InstallmentPlan, SubscriptionPeriod } from "./Subscriptions";
+import { TransactionTokenType, UsageLimit } from "./TransactionTokens";
+
+type BaseMetadata = Record<string, string>;
+
+export enum CheckoutType {
+    PAYMENT = "payment",
+    TOKEN = "token",
+}
+
+export enum ECFormCustomFieldType {
+    STRING = "string",
+    SELECT = "select",
+}
+
+export type ECFormCustomField = {
+    key: string;
+    label: string;
+    type: ECFormCustomFieldType;
+    required?: boolean;
+    options?: string[];
+};
+
+export type ECFormItem<Metadata = BaseMetadata> = {
+    /* EC Form data */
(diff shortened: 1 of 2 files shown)


src/resources/ECForms.ts
```diff
@@ -0,0 +1,121 @@
...
+    failureRedirectUrl?: string;
+
+    /* Charge and token data */
+    appId: string;
+    checkout: CheckoutType;
+    paymentType: PaymentShippingType;
+    tokenType: TransactionTokenType;
+    univapayCustomerId: string | null;
+    capture: boolean;
+    captureAt?: string | null;
```
laxa88, line 50: this `univapayCustomerId: string | null`, it is `null`-able. Does that mean this is optional?

src/resources/ECForms.ts
```diff
@@ -0,0 +1,121 @@
...
+    email?: string | null;
+    shippingAddressLine1: string | null;
+    shippingAddressLine2: string | null;
+    shippingAddressCity: string | null;
+    shippingAddressState: string | null;
+    shippingAddressZip: string | null;
+    shippingAddressCountryCode: string | null;
+    buyerName: string | null;
+    buyerNameTransliteration: string | null;
+    buyerDateOfBirth: string | null;
```
laxa88, line 97: what's the difference between the usage of `xxx?: type` and `xxx: type | null`?

laxa88 (approved):

src/resources/ECForms.ts
phieronymus: Here is a list of the different combination 🙂 

```tsx
required: Type; // Type
optional?: Type; // Type | undefined
nullable: Type | null // Type | null
optionalNullable?: Type | null // Type | null | undefind
```

`null` means that the property can be `null` and `undefined` means that the property may be `undefined` or not in the object.

src/resources/ECForms.ts
phieronymus: `univapayCustomerId: string | null,` means that we expect `univapayCustomerId` to always be a property in the object that can be a string or `null` when it does not exist 🙂 

I may review the parameters in another PR to remove the one we wont use in a first time. Just wanted to keep it simple for this PR as this was moved from the private SDK.

perrin4869 (approved):

perrin4869:
Looking good to me! maybe cool to reuse this type in the API side to keep them consistent. would need some typescript magic like `Omit` etc I think

phieronymus:
Yes. Let me do some cleanup in the next version to have the type we support only 🙂

PR #86, opened 2020-11-30
Rempve lodash

### LINKED ISSUE #76: Check for string type in error message for type safety

phieronymus:
Closes #76 

`lodash` was a devDependency that got used as a dependency therefore causing issues in consumers that are not using it. The safest solution is to remove lodash to let the consumer decide which library to use and prevent further confusion concerning the use of this library in the public SDK.


laxa88 (approved):

src/errors/RequestResponseError.ts
```diff
@@ -2,16 +2,14 @@
...
-import { isObject } from "lodash";
-
 import { ErrorResponse } from "../api/RestAPI";
 
 type ErrorRequest = Omit<ErrorResponse, "status">;
 
 const serializeErrorResponse = ({ code, httpCode, errors }: ErrorRequest): string =>
     `Code: ${code}, HttpCode: ${httpCode}, Errors: ${(errors || [])
         .map((error) =>
-            isObject(error) ? ("field" in error ? `${error.reason} (${error.field})` : error.reason) : error
```
perrin4869, line 14: Maybe it's better to reverse the line here?
```
typeof error === "string" ? error : ("field" in error ? `${error.reason} (${error.field})` : error.reason)
```

perrin4869 (commented):
Just a small suggestion

perrin4869 (approved):

src/errors/RequestResponseError.ts
phieronymus: Hum... That sounds reasonable to prevent the `undefined` and `null` case. Let me make the checks a bit more solid.

laxa88: if it's a number or array, then `error.reason` will throw

perrin4869:
this closes #76 too

src/errors/RequestResponseError.ts
perrin4869:
```javascript
> const foo = 10
undefined
> foo
10
> foo.reason
undefined
```

laxa88: ah sorry i didn't notice the `"field" in error` part, ignore my comment above 🙇

perrin4869: yeah so maybe `["number", "string"].indexOf(typeof error) > -1` is the better check

phieronymus:
How indeed! Thanks 🙂

phieronymus:
Also took the freedom of removing the nested ternary operator.

src/errors/RequestResponseError.ts
laxa88: indeed, and i forgot `typeof null` is actually `object`...

perrin4869:
Cool as long as error is not a number!

src/errors/RequestResponseError.ts
phieronymus: `null` and `undefined` should be filterd now so it wont be crashing in this case. Let me log the error for `[string, number, boolean]`. Thanks for noticing it. The types being wrong here we can expect anything... We may try to fix them at one point to know what to expect though.

perrin4869:
Wouldn't this be nice here? XD
https://github.com/tc39/proposal-pattern-matching

src/errors/RequestResponseError.ts
```diff
@@ -15,7 +15,7 @@ const serializeErrorResponse = ({ code, httpCode, errors }: ErrorRequest): strin
                 return error; // (pH) TODO: special handling for wrongly formatted. Fix the ErrorRequest type.
 
             default:
-                return `${error.reason}${"field" in error && ` (${error.field})`}`;
+                return "field" in error ? `${error.reason} (${error.field})` : `${error.reason}`;
```
laxa88, line 18: this will still throw if `error` is null

PR #92, opened 2020-12-02
Fix/pierreh/integrate ec form api

phieronymus:
Updates https://github.com/univapaycast/univapay-form-checkout/issues/51

### DIFF
--- a/src/errors/RequestResponseError.ts
+++ b/src/errors/RequestResponseError.ts
@@ -7,12 +7,13 @@ import { ErrorResponse } from "../api/RestAPI";
 type ErrorRequest = Omit<ErrorResponse, "status">;
 
 const serializeErrorResponse = ({ code, httpCode, errors }: ErrorRequest): string => {
-    const formattedErrors = (errors || []).filter(Boolean).map((error) => {
+    // (pH) FIXME: `errors sometimes is not an array. Fix type and maybe the cause of it
+    const formattedErrors = (errors || [])?.filter(Boolean).map((error) => {
         switch (typeof error) {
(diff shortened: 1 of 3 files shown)


src/resources/ECForms.ts
```diff
@@ -108,14 +108,15 @@ export type ResponseECForm<Metadata = BaseMetadata> = ECFormItem<Metadata>;
 
 export class ECForms extends CRUDResource {
     static requiredParams: string[] = [];
-    static routeBase = "/forms";
+    static routeBase = "/merchants/:merchantId/forms";
```
phieronymus, line 111: We do not have authorization headers in the form checkout at the point of querying this. Moving it in the parameters instead.

src/errors/RequestResponseError.ts
```diff
@@ -7,12 +7,13 @@ import { ErrorResponse } from "../api/RestAPI";
 type ErrorRequest = Omit<ErrorResponse, "status">;
 
 const serializeErrorResponse = ({ code, httpCode, errors }: ErrorRequest): string => {
-    const formattedErrors = (errors || []).filter(Boolean).map((error) => {
+    // (pH) FIXME: `errors sometimes is not an array. Fix type and maybe the cause of it
+    const formattedErrors = (errors || [])?.filter(Boolean).map((error) => {
```
phieronymus, line 11: Somehow the `errors` array sometimes is not an array... I will investigate the types of the errors at one point. I think it became more and more messy and the type was not updated accordingly.

laxa88 (approved):

PR #106, opened 2020-12-17
Fix/pierreh/update dependencies

phieronymus:
No ticket linked to this issue:

- Update dependencies
- Revert string-query bump (current version is not working on IE11 and edge (even with compilation and polyfills))
- Use another casing library as the current ones are not working on IE11 and edge (even with compilation and polyfills)


test/specs/payload.specs.ts
```diff
@@ -58,7 +57,7 @@ describe("Payload Helpers", () => {
                 {
                     foo: "bar",
                     foo2: {
-                        bar: new Buffer(["test"]),
+                        bar: Buffer.from("test"),
```
phieronymus, line 60: `new Buffer` is deprecated. Not an important change though. It should not impact the tests.

test/specs/payload.specs.ts
```diff
@@ -81,11 +80,12 @@ describe("Payload Helpers", () => {
             },
             foo2: ["bar"],
             foo3: [{ bar: "fizz" }],
-            foo4: new Buffer(["test"]),
+            foo4: Buffer.from("test"),
             foo5: undefined,
             foo6: {
                 bar: undefined,
             },
+            foo7: [{ fooBar: "fizz" }],
```
phieronymus, line 88: Added one more test case

src/utils/object.ts
```diff
@@ -5,35 +5,29 @@
...
 
+    const ignoredKeySet = new Set(ignoreKeys);
     return Object.keys(obj || {}).reduce((acc: Record<string, any>, key: string) => {
-        let value = obj[key];
+        const formattedKey = ignoredKeySet.has(key) ? key : transformer(key);
+        const value = isObject(obj[key]) ? transformKeys(obj[key], transformer) : obj[key];
 
-        if (ignoreKeys.includes(key)) {
-            return { ...acc, [key]: value };
-        }
```
phieronymus, line 23: Ups... Let me update the ignored key set again.

phieronymus: Ok me-proofed-tests added, we should be fine now 🙂

package.json
```diff
@@ -61,29 +61,24 @@
...
-    "uuid": "8.3.1"
+    "typescript": "4.1.3",
+    "uuid": "8.3.2"
   },
   "dependencies": {
-    "camelcase": "^6.2.0",
-    "decamelize": "^4.0.0",
+    "change-case": "^4.1.2",
     "events": "^3.2.0",
-    "flat": "^5.0.2",
```
phieronymus, line 77: This library is not used.

src/api/utils/payload.ts
```diff
@@ -24,52 +26,35 @@ export function containsBinaryData(data: any): boolean {
...
-                }
+    const appendFormData = (data: unknown, root = "", key = "") => {
+        const formattedKey = keyFormatter(key);
+        const path = !root ? formattedKey : formattedKey ? `${root}.${formattedKey}` : root;
+
+        if (isBlob(data)) {
+            // Blob
+            formData.append(path, data);
+        } else if (Array.isArray(data)) {
+            // Array
```
laxa88, line 42: unnecessary comments? 🤔

src/utils/object.ts
```diff
@@ -5,35 +5,33 @@
...
 
-        if (isObject(value)) {
-            if (Array.isArray(value)) {
-                value = transformArray(value);
-            } else {
-                value = transformKeys(value, transformer);
-            }
-        }
-
-        acc[transformer(key)] = value;
```
laxa88, line 33: just wondering, the logic changed here a little, is it an improvement or bugfix? 🤔

laxa88 (approved):

src/api/utils/payload.ts
phieronymus: Hum... I jut tried to list the cases as it was not obvious that the last one was for string and buffers. I can remove the comments if you think it is extra though 🙂

src/utils/object.ts
phieronymus: The function already handles the arrays 🙂 Which is why the logic is still the same. The array handling was repeated.

Also made the code immutable in this function (except for `acc` as per convention)

phieronymus: Changed the `obj[key]` back to `value` 🙂 Thanks!

phieronymus: The linked test to my first point is the array test:

```
myKey2: [
                {
                    mySubKey21: "Value 4-1",
                    mySubKey22: "Value 4-2",
                    mySubKey23: ["Value 4-3-1", "Value 4-3-2"],
                },
            ],
```

laxa88 (approved):

src/api/utils/payload.ts
laxa88: doesn't hurt to keep the comments :meow_party:

PR #108, opened 2020-12-21
Fix/pierreh/fix nested ignored keys

phieronymus:
Fix for an issue caused by the changes for the IE/Edge compatibility

The code was relying heavily on the fact that the old library was only changing the `camelCase`, ignoring that metadata can have keys formatted in camelCase as well.

The PR changes three things:
- Use `transformKeys` in the `objectToFormData` function as they are supposed to be the same formatting
- Fix nested ignoredKeys in `transformKeys`
- Add the `metadata` to the ingoredKeys for the `getRequestBody` function as it needed here as well

### DIFF
--- a/test/specs/object.specs.ts
+++ b/test/specs/object.specs.ts
@@ -21,10 +21,11 @@ describe("Object Helpers", () => {
                 },
             ],
             myKey3: ["Value 5-1", "Value 5-2"],
-            myIgnoredKey1: { myIgnoredKey11: "Ignored value 1", "my-ignored-key=12": "Ignored value 2" },
+            myIgnoredKey1: { myIgnoredKey11: "Ignored value 11", "my-ignored-key=12": "Ignored value 12" },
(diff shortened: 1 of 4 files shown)


test/specs/object.specs.ts
```diff
@@ -44,8 +45,11 @@ describe("Object Helpers", () => {
             ],
             my_key3: ["Value 5-1", "Value 5-2"],
             myIgnoredKey1: {
-                myIgnoredKey11: "Ignored value 1",
-                "my-ignored-key=12": "Ignored value 2",
+                myIgnoredKey11: "Ignored value 11",
+                "my-ignored-key=12": "Ignored value 12",
```
phieronymus, line 49: Oh wait... Let me review this.

phieronymus: Ah no... Git diff made over-react

laxa88 (approved):

PR #109, opened 2020-12-22
Change jwt-decode to es5 compatible library

phieronymus:
No ticket linked to this issue. Fixing the es5 compatibility for ie.


phieronymus:
It should be the last package we directly import that is not es5 compatible. Hopefully this is the last one for this project 🙂 

Used the `are-you-es5 check` library to check. Note that this library does not check dependencies of dependencies if they are not bundled with the package so it is not infailable.

laxa88 (approved):

PR #110, opened 2020-12-22
Update ECForms types to allow null values

perrin4869:
hm... something like this? Optional fields are kinda hard to handle on my end, nullable fields are easier!

### DIFF
--- a/src/resources/ECForms.ts
+++ b/src/resources/ECForms.ts
@@ -38,18 +38,19 @@ export type ECFormItem<Metadata = BaseMetadata> = {
     createdOn: string;
 
     /* Redirects */
-    successRedirectUrl?: string;
-    failureRedirectUrl?: string;
+    successRedirectUrl: string | null;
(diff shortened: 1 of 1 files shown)


perrin4869:
hm... depends what `ECForm` will be used for I guess - if it's for `POST` operations, optional fields are cool

perrin4869:
Had a change of heart and returned the `?` to the fields - you can still send me `POST` and `PATCH` forms without those fields, but I'll default them all to null - they will be present in all `GET` requests

perrin4869:
Guess the whole point here is that the fields are nullable lol

phieronymus:
We may think of splitting the types at one point for query and mutation. For now I believe this is good for the current usecase and do a big refactoring seems ta bit out of the scope of this PR.

Let me merge this 🙂 Also thanks for adding the pendingRedirectURL

phieronymus (approved):

PR #113, opened 2021-01-08
Fix/pierreh/fix todos

phieronymus:
No ticket associated to this PR.

The only TODO remaining is the test todo for form data. If handled it would be in another Pull Request as investigation is needed. Other todos/fixmes have been fixed in this Pull Request

### DIFF
--- a/src/errors/RequestResponseError.ts
+++ b/src/errors/RequestResponseError.ts
@@ -4,27 +4,56 @@
 
 import { ErrorResponse, SubError, ValidationError } from "../api/RestAPI";
 
-type ErrorRequest = Omit<ErrorResponse, "status">;
+import { RequestErrorCode, ResponseErrorCode } from "./APIError";
 
-const isSymbol = (value: unknown): value is symbol => ["string", "number", "boolean"].includes(typeof value);
-const serializeErrorResponse = ({ code, httpCode, errors }: ErrorRequest): string => {
-    // (pH) FIXME: `errors sometimes is not an array. Fix type and maybe the cause of it
-    const formattedErrors = (() => {
-        if (!errors || isSymbol(errors)) {
-            return errors;
+type ErrorItem = boolean | number | string | SubError | ValidationError;
+type RawErrorRequest = {
+    httpCode?: number;
+    code: ResponseErrorCode | RequestErrorCode;
+    errors: ErrorItem | ErrorItem[];
+};
+
+type FormattedErrorRequest = Omit<ErrorResponse, "status">;
+
+const isSymbol = (value: unknown): value is string | number | boolean =>
+    ["string", "number", "boolean"].includes(typeof value);
+
+/**
+ * The function format the ill-formatted errors:
+ * - When errors is not an array, use it as the only error
+ * - Filters null and undefined errors
+ * - When an item of errors is a symbol rather than an object, returns the UNKNOWN_ERROR as the code
+ */
+const formatErrors = ({ errors: rawErrors, ...rest }: RawErrorRequest): FormattedErrorRequest => {
+    const errorList = Array.isArray(rawErrors) ? rawErrors : [rawErrors];
+
+    const formattedErrors: (SubError | ValidationError)[] = errorList.filter(Boolean).map((error: ErrorItem) => {
+        if (typeof error === "string") {
+            const errorCodeKeys = [...Object.keys(ResponseErrorCode), ...Object.keys(RequestErrorCode)];
+            const isValidError = errorCodeKeys.some(
+                (key) => ResponseErrorCode[key] === error || RequestErrorCode[key] === error
+            );
+
+            return {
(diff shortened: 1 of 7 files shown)


src/errors/RequestResponseError.ts
```diff
@@ -4,50 +4,81 @@
...
-        super(serializeErrorResponse(errorResponse));
+    constructor(errorResponse: RawErrorRequest) {
+        const formattedErrorResponse = formatErrors(errorResponse);
+        super(serializeErrorResponse(formattedErrorResponse));
+
         Object.setPrototypeOf(this, RequestResponseBaseError.prototype);
-        this.errorResponse = { status: "error", ...errorResponse };
+        this.errorResponse = { status: "error", ...formattedErrorResponse };
     }
 }
```
phieronymus, line 71: Ok, wha changed here is that the error is formatted to fit the right type. Before the `serializeErrorResponse` was handling this but the error inside the `RequestResponseBaseError` still was ill-formatted that can cause issues in the consumers when using the object as the type was wrong.

Now it should be consistent with the information typescript is giving.

src/utils/object.ts
```diff
@@ -66,9 +66,3 @@ export function missingKeys(obj: Record<string, any>, keys: string[] = []): stri
 
     return missing;
 }
-
-export function omit<T extends Record<string, unknown>, K extends keyof T>(obj: T, keys: K[]): Omit<T, K> {
```
phieronymus, line 70: This function was not needed anymore.

src/errors/RequestResponseError.ts
phieronymus: Note that more tests have been added here as well.

test/specs/errors.specs.ts
```diff
@@ -33,4 +33,51 @@ describe("Errors", () => {
             "Error: Code: REQUEST_ERROR, HttpCode: 400, Errors: REQUEST_ERROR, REQUEST_ERROR (my no reason field)"
         );
     });
+
+    it("should serialize the ResponseError correctly", async () => {
```
laxa88, line 37: This description is duplicated. 🤔

test/specs/errors.specs.ts
```diff
@@ -33,4 +33,51 @@ describe("Errors", () => {
...
+                { reason: ResponseErrorCode.InvalidFormat, field: "dummy_field" },
+            ],
+        });
+
+        expect(error.toString()).to.equal(
+            "Error: Code: VALIDATION_ERROR, HttpCode: 400, Errors: INVALID_AMOUNT, INVALID_FORMAT (dummy_field)"
+        );
+    });
+
+    it("should serialize an error with `errors` not being an array", async () => {
```
laxa88, line 52: probably can shorten this to `should serialize non-array errors`

test/specs/errors.specs.ts
```diff
@@ -33,4 +33,51 @@ describe("Errors", () => {
...
+        const error = new RequestError({
+            httpCode: 400,
+            code: RequestErrorCode.RequestError,
+            errors: { reason: RequestErrorCode.RequestError },
+        });
+
+        expect(error.toString()).to.equal("Error: Code: REQUEST_ERROR, HttpCode: 400, Errors: REQUEST_ERROR");
+    });
+
+    it("should serialize an error with `errors` using a symbol as error item", async () => {
```
laxa88, line 62: probably can shorten this to `should serialize symbol errors`

test/specs/errors.specs.ts
```diff
@@ -33,4 +33,51 @@ describe("Errors", () => {
...
+            code: RequestErrorCode.RequestError,
+            errors: [RequestErrorCode.RequestError, "invalid code", 1, true],
+        });
+
+        expect(error.toString()).to.equal(
+            "Error: Code: REQUEST_ERROR, HttpCode: 400, Errors: REQUEST_ERROR, UNKNOWN_ERROR, UNKNOWN_ERROR, UNKNOWN_ERROR"
+        );
+    });
+
+    it("should serialize an error with `errors` having null and undefined values", async () => {
```
laxa88, line 74: probably can shorten this to `should serialize undefined and null errors`

laxa88 (approved):

test/specs/errors.specs.ts
phieronymus: Actually it would be confusing as the `error` object contains the `errors` object/array. Not specifying will make it harder to know which `error` we speak about 🤔

phieronymus: Same for the two next comments. I believe that shortening it will introduce some understanding issue as the object is structured in a way which is prone to misunderstanding when describing.

I would feel better being a bit verbose to describe exactly what we want. But I may be missing something here.

test/specs/errors.specs.ts
phieronymus: Ups. Seems that we already had the test. Thanks 🙂

PR #118, opened 2021-01-13
Replace isomorphic-fetch with cross-fetch

perrin4869:
This thing drove me crazy lol
Working on updates to the miniapp api, I noticed that requests to gpp stopped working with newer `univapay-node` versions.
After lots of digging, found that the cause was the upgrade to `isomorphic-fetch` [here](https://github.com/univapay/univapay-node/commit/0b78e088fbe0892f260b474dd6e6f49f10701fae).
`isomorphic-fetch@3` uses `node-fetch@2`, which introduced an `esm` module. When bundling with rollup, it would import `node-fetch` as a namespace, but `isomorphic-fetch` treated it as a cjs import. Long story short, everything beaks.
There is a fix pending in `isomorphic-fetch` https://github.com/matthew-andrews/isomorphic-fetch/pull/195, but doesn't seem like it'll be merged soon. On the other hand, `cross-fetch` already has the fix merged in and seems better maintained...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1530,6 +1530,14 @@
       "integrity": "sha512-dcKFX3jn0MpIaXjisoRvexIJVEKzaq7z2rZKxf+MSr9TkdmHmsU4m2lcLojrj/FHl8mk5VxMmYA+ftRkP/3oKQ==",
       "dev": true
     },
+    "cross-fetch": {
+      "version": "3.0.6",
+      "resolved": "https://registry.npmjs.org/cross-fetch/-/cross-fetch-3.0.6.tgz",
+      "integrity": "sha512-KBPUbqgFjzWlVcURG+Svp9TlhA5uliYtiNx/0r8nv0pdypeQCRJ9IaSIc3q/x3q8t3F75cHuwxVql1HFGHCNJQ==",
+      "requires": {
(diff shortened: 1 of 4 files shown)


perrin4869:
The other option, which I think I prefer, is to leave out isomorphic-fetch entirely, and leave it to the end users to polyfill fetch if they need to. If we go with that option, we need to document in the README which polyfills will be necessary.

phieronymus:
Sorry for that. Seems that I am the one who broke this up. I did not notice the two versions in the package-lock.

To be honest isomorphic-fetch has not been updated for a few years now. I believe that changing the package is a good idea.

I also believe that letting the consumer polyfill would be a good idea though. We do not have this issue on our application but some user may want to use other polyfill that work with their own sdk as well 🤔

perrin4869:
The other thing is that many applications don't even need to polyfill - if the app is targeting modern browsers then isomorphic-fetch is being wasted...

perrin4869:
Oh and don't worry, this was a very specific problem that wouldn't be noticed unless you're having a setup like in the services-typescript repo :P

laxa88 (approved):
LGTM

perrin4869:
OK, let's merge this in for now, and think about what to do about polyfilling before v1

PR #123, opened 2021-02-09
Upgrade to husky@5

perrin4869:
Just upgraded in `services-typescript` so might as well upgrade everywhere else XD

https://dev.to/typicode/what-s-new-in-husky-5-32g5


laxa88:
just wondering, what's pinst for?

laxa88 (approved):

phieronymus:
@laxa88 It is used in the package.json to control whether the hooks are enabled or not on certain commands 🙂 

```
    "prepublishOnly": "pinst --disable",
    "postpublish": "pinst --enable"
```

laxa88:
@phieronymus i know what it does, but ... i don't understand why we're using it? 🙇

perrin4869:
Took it straight from the documentation. It makes sure that when you install this module as a dependency, husky install isn't run

PR #127, opened 2021-02-16
Update dev dependencies

phieronymus:
Update dev dependencies to fix the dependabot alert:

```
 Dependabot cannot update marked to a non-vulnerable version
The latest possible version that can be installed is 1.2.9 because of the following conflicting dependency:

typedoc@0.20.19 requires marked@^1.2.5
The earliest fixed version is 2.0.0.
```


package.json
```diff
@@ -69,9 +69,9 @@
     "sinon": "9.2.4",
     "sinon-chai": "3.5.0",
     "ts-node": "9.1.1",
-    "typedoc": "0.20.19",
+    "typedoc": "0.20.25",
```
phieronymus, line 72: This is the minimal change to be done. The rest is maintenance.

laxa88 (approved):

PR #131, opened 2021-02-23
Remove polyfills

perrin4869:
Been having issues using the SDK on `lambda` again, this time related to the polyfills I believe (I think it is related to [this issue](https://github.com/node-fetch/node-fetch/issues/1015)).
I'm thinking to avoid these issues it might be best to leave polyfilling up to the clients. Also, I think this could be enough to cut `v1.0` and make it easier to keep it in sync with the private SDK.


phieronymus (approved):
This makes sense actually. If this causes problem on top of it there is not reason to be against this change 🙂 Thanks!

PR #135, opened 2021-03-11
Use is-buffer instead of using Buffer object

phieronymus:
Remove unnecessary dependency on the node.js Buffer. This is now using a lighter object that does not need polyfills

### DIFF
--- a/src/utils/object.ts
+++ b/src/utils/object.ts
@@ -4,6 +4,7 @@
  */
 
 import { camelCase, snakeCase } from "change-case";
+import isBuffer from "is-buffer";
 
 export type Transformer = (...args: any[]) => string;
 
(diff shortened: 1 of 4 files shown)


src/utils/object.ts
```diff
@@ -24,11 +25,11 @@ export const transformKeys = (
     transformer: Transformer,
     ignoreKeys: string[] = []
 ): any => {
-    const transformArray = (arr: unknown[], transformer: Transformer): any =>
+    const transformArray = (arr: unknown[]): any =>
         arr.map((item) => (isObject(item) ? transformKeys(item, transformer, ignoreKeys) : item));
 
     if (Array.isArray(obj)) {
-        return transformArray(obj, transformer);
+        return transformArray(obj);
```
phieronymus, line 32: Yeah... This is not really related but there was a shadow variable... Better fix that now 🙂

laxa88 (approved):

PR #137, opened 2021-03-18
Add tagged credentials gateway error

phieronymus:
Updates https://github.com/univapaycast/univapay-admin-console/issues/1143

Adding this for the gateway simulation as well 🙂

### DIFF
--- a/src/errors/APIError.ts
+++ b/src/errors/APIError.ts
@@ -221,7 +221,7 @@ export enum ResponseErrorCode {
 
     /* Gateway simulation */
     PlatformCredentialsDisabled = "PLATFORM_CREDENTIALS_DISABLED",
-    TaggedPlatformCredentialDisabled = "TAGGED_PLATFORM_CREDENTIAL_DISABLED",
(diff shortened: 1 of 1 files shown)


laxa88 (approved):

src/errors/APIError.ts
```diff
@@ -221,6 +221,7 @@ export enum ResponseErrorCode {
 
     /* Gateway simulation */
     PlatformCredentialsDisabled = "PLATFORM_CREDENTIALS_DISABLED",
+    TaggedPlatformCredentialsDisabled = "TAGGED_PLATFORM_CREDENTIAL_DISABLED",
```
laxa88, line 224: perhaps make this consistent? minor typo on the `CREDENTIAL` > `CREDENTIALS`

phieronymus: Ah thanks 🙂 For now the error is with the typo. I will notify the API about it and update it in a later PR. 🙂

phieronymus: Also added another message I happened to see.

phieronymus: Hum... After thinking about it. Fixed it in here. This wont break and it will be ready for the fixed error. Thanks for catching it! 🙂

PR #142, opened 2021-04-22
Add missing brands and gateways

phieronymus:
Some brands were missing.. Also re-adding the aliapy_connect_mpm I removed as it actually was valid.

### DIFF
--- a/src/resources/common/enums.ts
+++ b/src/resources/common/enums.ts
@@ -42,7 +42,7 @@ export enum CardCategory {
     ELECTRON = "electron",
     MAESTRO = "maestro",
     WORLD = "world",
-    BUSINESS ="business",
+    BUSINESS = "business",
 }
 
 export enum ProcessingMode {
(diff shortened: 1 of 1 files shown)


src/resources/common/enums.ts
```diff
@@ -77,6 +78,7 @@ export enum QRBrand {
     OKI_PAY = "oki_pay",
     ORIGAMI = "origami",
     PAY_PAY = "pay_pay",
+    PAY_PAY_MERCHANT = "pay_pay_merchant",
```
phieronymus, line 81: This one was missing somehow. Adding it at the same time.

src/resources/common/enums.ts
```diff
@@ -65,6 +65,7 @@ export enum QRBrand {
     D_BARAI = "d_barai",
     EZLINK = "ezlink",
     GCASH = "gcash",
+    GINKO_PAY = "ginko_pay",
```
phieronymus, line 68: Ginko pay is a gateway but also a brand. Adding it as it was missing.

laxa88 (approved):

PR #146, opened 2021-06-28
Update package

phieronymus:
Just update the libraries. Two libraries were not updated:

- `query-string` (as updating it would break IE11 support)
- `ts-node` as the esm features needed for the tests are not fully supported yet.


phieronymus:
My bad, seems that I did something wrong. Let me fix 🙂

laxa88 (approved):

PR #147, opened 2021-06-28
Build esm with extensions

phieronymus:
After patching the consoles (which may be useful in some cases), here is the sdk fix to add the extensions to esm. That should make it backward compatible 🙂

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,11 +72,11 @@
         "sinon": "11.1.1",
         "sinon-chai": "3.7.0",
         "ts-node": "9.1.1",
-        "ttypescript": "^1.5.12",
+        "ttypescript": "1.5.12",
         "typedoc": "0.21.2",
         "typedoc-plugin-internal-external": "2.2.0",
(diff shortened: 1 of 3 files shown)


5ym (approved):

package.json
```diff
@@ -72,9 +72,11 @@
         "sinon": "11.1.1",
         "sinon-chai": "3.7.0",
         "ts-node": "9.1.1",
+        "ttypescript": "^1.5.12",
         "typedoc": "0.21.2",
         "typedoc-plugin-internal-external": "2.2.0",
         "typescript": "4.3.4",
+        "typescript-transform-extensions": "^1.0.1",
```
laxa88, line 79: should we pin these?

laxa88 (approved):
thank you! 🙇‍♂️

perrin4869 (approved):
Nice!

perrin4869:
btw, package-lock.json should be responsible for pinning the deps I think

PR #153, opened 2021-08-11
add interval for long poll

laxa88:
No associated ticket. Our `longPoll` doesn't actually have intervals so it actually rapid-fires the endpoint until the desired result is achieved... which may not be a good idea for the API. 🤔

### DIFF
--- a/src/api/RestAPI.ts
+++ b/src/api/RestAPI.ts
@@ -12,6 +12,7 @@ import {
     ENV_KEY_ENDPOINT,
     ENV_KEY_SECRET,
     IDEMPOTENCY_KEY_HEADER,
+    POLLING_INTERVAL,
     POLLING_TIMEOUT,
 } from "../common/constants";
 import { RequestErrorCode, ResponseErrorCode } from "../errors/APIError";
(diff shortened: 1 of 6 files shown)


jdebono (approved):

src/api/RestAPI.ts
```diff
@@ -312,7 +314,12 @@ export class RestAPI extends EventEmitter {
                 }),
                 // Repeater
                 (async function repeater(): Promise<A> {
-                    const result = await promise();
+                    const result = await new Promise<A>((resolve) => {
+                        setTimeout(async () => {
+                            const r = await promise();
+                            resolve(r);
+                        }, interval);
+                    });
```
phieronymus, line 322: Not important, feel free to skip this comment.

This could be simplified to:

```tsx
await new Promise(resolve => setTimeout(resolve, interval)); // sleep to avoid firing requests too fast
```

Or am I missing something? 🙂

phieronymus: Also it may be overkilling it but we may want to add a flag in `execRequest` to know if this is the first time and do the first trigger immediately. Quite a lot of poll end up only doing one call on the transaction history list and we do not want to make it slower in those cases 🙂

phieronymus (approved):

src/api/RestAPI.ts
laxa88: let me see if i can refactor the execRequest logic too

PR #167, opened 2021-10-26
Migrate to github actions

perrin4869:
Change of pace PR
We need to add `NPM_TOKEN` as an organization secret to publish package, but this should be good otherwise


phieronymus:
Merging this PR. Thanks a lot for the update 🙂

PR #192, opened 2021-11-10
Fix/pierreh/fix metadata bigint

phieronymus:
For big int in the metadata the frontend works in an unexpected manner. This fixes most of the issues related to it.

We will still have to make some modification to the consoles to ensure the metadata are presented in the correct way.

Also removed a `any` as I do not like them and added some free testing.

### DIFF
--- a/src/api/RestAPI.ts
+++ b/src/api/RestAPI.ts
@@ -97,19 +97,18 @@ export type PromiseCreator<A> = () => Promise<A>;
 
 export type SendData<Data> = Data;
 
-function getRequestBody<Data>(data: SendData<Data>, keyFormatter = toSnakeCase): string | FormData {
-    return containsBinaryData(data)
+const getRequestBody = <Data>(data: SendData<Data>, keyFormatter = toSnakeCase): string | FormData =>
+    containsBinaryData(data)
         ? objectToFormData(data, keyFormatter, ["metadata"])
         : JSON.stringify(transformKeys(data, keyFormatter, ["metadata"]));
-}
 
-function stringifyParams<Data extends Record<string, any>>(data: Data): string {
+const stringifyParams = <Data extends Record<string, any>>(data: Data): string => {
     const query = stringifyQuery(transformKeys(data, toSnakeCase), { arrayFormat: "bracket" });
 
     return query ? `?${query}` : "";
-}
+};
 
-async function execRequest<A>(executor: () => Promise<A>, callback?: ResponseCallback<A>): Promise<A> {
+const execRequest = async <A>(executor: () => Promise<A>, callback?: ResponseCallback<A>): Promise<A> => {
     try {
         const response = await executor();
         if (typeof callback === "function") {
@@ -125,7 +124,7 @@ async function execRequest<A>(executor: () => Promise<A>, callback?: ResponseCal
 
         throw err;
     }
-}
+};
 
 export class RestAPI extends EventEmitter {
     endpoint: string;
@@ -175,16 +174,16 @@ export class RestAPI extends EventEmitter {
     /**
      * @internal
      */
-    async send<A, Data = any>(
+    async send<ResponseBody, Data = unknown>(
         method: HTTPMethod,
         uri: string,
         data?: SendData<Data>,
         auth?: AuthParams,
-        callback?: ResponseCallback<A>,
+        callback?: ResponseCallback<ResponseBody>,
         requireAuth = true,
         acceptType?: string,
         keyFormatter = toSnakeCase
-    ): Promise<A> {
+    ): Promise<ResponseBody | string | Blob | FormData> {
         const dateNow = new Date();
         const timestampUTC = Math.round(dateNow.getTime() / 1000);
 
@@ -209,7 +208,7 @@ export class RestAPI extends EventEmitter {
 
         this.emit("request", request);
 
-        return execRequest(async () => {
+        return execRequest<ResponseBody | string | Blob | FormData>(async () => {
             const response = await fetch(request);
 
             this.emit("response", response);
@@ -230,7 +229,7 @@ export class RestAPI extends EventEmitter {
 
             const contentType = response.headers.get("content-type");
             if (contentType === "application/json") {
-                return parseJSON(response, ["metadata"]);
+                return parseJSON<ResponseBody>(response, ["metadata"]);
             } else if (contentType) {
                 if (contentType.indexOf("text/") === 0) {
                     return response.text();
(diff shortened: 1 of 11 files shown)


src/resources/Resource.ts
```diff
@@ -99,7 +99,8 @@ export abstract class Resource extends EventEmitter {
                 return Promise.reject(err);
             }
 
-            return api.send(method, url, data, auth, callback, requireAuth, acceptType, keyFormatter);
+            // TODO: Have a look as this cast, the response can be other objects such as string, Blob or FormData
+            return api.send(method, url, data, auth, callback, requireAuth, acceptType, keyFormatter) as Promise<B>;
```
phieronymus, line 103: It was following the type previously only because of the `any` that was screwing up all the typing until there

src/api/RestAPI.ts
```diff
@@ -175,16 +175,16 @@ export class RestAPI extends EventEmitter {
...
         uri: string,
         data?: SendData<Data>,
         auth?: AuthParams,
-        callback?: ResponseCallback<A>,
+        callback?: ResponseCallback<ResponseBody>,
         requireAuth = true,
         acceptType?: string,
         keyFormatter = toSnakeCase
-    ): Promise<A> {
+    ): Promise<ResponseBody | string | Blob | FormData> {
```
phieronymus, line 187: This is way more accurate than it was before.

src/utils/object.ts
```diff
@@ -32,7 +32,7 @@ export const transformKeys = (
         return transformArray(obj);
     }
 
-    const ignoredKeySet = new Set(ignoreKeys || []);
+    const ignoredKeySet = new Set(ignoreKeys);
```
phieronymus, line 35: This is already handled in the default assignment. `null` is not a valid value here.

test/specs/sdk.specs.ts
```diff
@@ -50,4 +50,25 @@ describe("SDK", () => {
                 .equal(sdk.api);
         }
     });
+
+    it("should add listener on the sdk", () => {
```
phieronymus, line 54: Not really related, just added those tests as I wanted to know if some branches tests were missing

test/specs/fetch.specs.ts
```diff
@@ -0,0 +1,63 @@
...
+
+const createResponse = (data: Record<string, unknown>, status = 200) =>
+    new Response(JSONBig({ useNativeBigInt: true }).stringify(data, null, 2), {
+        "status": status,
+        "statusText": "Dummy status",
+    });
+
+describe("Fetch Helpers", () => {
+    describe("parseJSON", () => {
+        it("return the parsed response body", async () => {
```
perrin4869, line 15: can you clarify what this is testing?

test/specs/fetch.specs.ts
```diff
@@ -0,0 +1,63 @@
...
+        });
+
+        it("return the parsed response body without formatting the keys in parameter", async () => {
+            const result = await parseJSON(createResponse({ key1: "1", key2: "2", key3: { key_31: 1 } }), ["key3"]);
+
+            expect(result).eql({ key1: "1", key2: "2", key3: { key_31: 1 } });
+        });
+
+        it("return the parsed response body with bigint", async () => {
+            const result = await parseJSON(createResponse({ key1: BigInt("0123456789101112131415") }), ["key3"]);
```
perrin4869, line 44: `123456789101112131415n` maybe looks better?

perrin4869 (commented):
Few comments but looks good!

test/specs/fetch.specs.ts
phieronymus: This is testing that the response body get parsed, which is what the function does. This is the general case to see if everything goes as expected when provided a response.

test/specs/sdk.specs.ts
```diff
@@ -50,4 +50,25 @@ describe("SDK", () => {
...
+        const sdk = new SDK({ endpoint: "/" });
+
+        const expectedPromise = new Promise((resolve) => sdk.addListener("test", () => resolve(true)));
+
+        sdk.emit("test");
+
+        expect(expectedPromise).to.eventually.be.fulfilled;
```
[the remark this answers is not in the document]
phieronymus: The test is pretty straight-forward. We do not use the p-event library otherwise so I do not know if we need to introduce this dependency for only that code. Thoughts?

src/utils/fetch.ts
```diff
@@ -3,20 +3,25 @@
...
 
-export async function parseJSON(response: Response, ignoreKeys: string[] = ["metadata"]): Promise<any> {
+export const parseJSON = async <FormattedBody>(
+    response: Response,
+    ignoreKeys: string[] = ["metadata"]
+): Promise<FormattedBody> => {
     const text = await response.text();
-    return text ? transformKeys(JSON.parse(text), toCamelCase, ignoreKeys) : {};
-}
+    return text ? transformKeys(JSONBig({ useNativeBigInt: true }).parse(text), toCamelCase, ignoreKeys) : {};
```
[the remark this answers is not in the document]
phieronymus: Sounds good, thanks 🙂

test/specs/fetch.specs.ts
phieronymus: Indeed, a lot nicer, thanks! 🙂

test/specs/sdk.specs.ts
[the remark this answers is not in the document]
perrin4869: yeah no need!

test/specs/fetch.specs.ts
laxa88: if you're asserting `parseJson` result, perhaps a unit test for `parseJson` would be sufficient? 🤔 and in this case here we could probably assert the stringified json instead 🤔

laxa88 (approved):

test/specs/fetch.specs.ts
phieronymus: Note that is is testing with a response. We do get a response in parameter and parse its body. The JSON.parse does not get the response body here. Some other things are happening in here as well. As this is a generic utils it is better to have basic tests here.

Maybe I do not understand what you mean though 🤔

laxa88: oohh... i see, my bad! 🙇

test/specs/fetch.specs.ts
phieronymus: Hum... Seems that node 12 does not really like the target being es2020 and ts wouldn't allow me to choose something lower that that. Since this is a minor change we may want to have it in another PR.

perrin4869:
ahhh that's a shame haha!

PR #193, opened 2021-11-11
Sanitise data object before sending in defineRoute

laxa88:
Updates https://github.com/univapaycast/univapay-finance/issues/1244

Not really related to above ticket, but I found this issue while working on that ticket. 🙇

### DIFF
--- a/src/resources/Resource.ts
+++ b/src/resources/Resource.ts
@@ -48,6 +48,24 @@ function compilePath(path: string, pathParams: Record<string, string>): string {
         .replace(/:(\w+)/gi, (s: string, p: string) => pathParams[p] || s);
 }
 
+type ObjectType =
+    | "array"
+    | "bigint"
+    | "boolean"
+    | "function"
+    | "null"
+    | "number"
+    | "object"
+    | "string"
+    | "symbol"
+    | "undefined";
+
+const getObjectType = (o: unknown): ObjectType => {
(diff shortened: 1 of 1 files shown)


src/resources/Resource.ts
```diff
@@ -76,11 +76,20 @@ export abstract class Resource extends EventEmitter {
...
             pathParams: Record<string, string> = {}
         ): Promise<B> {
+            /**
+             * Sanitises and ensures that the data is recreated as an object with default prototypes.
+             * In rare cases if the data was created with `Object.create(null)` (no prototypes),
+             * this will cause the data to be sent as `WebKitFormBoundary` instead of regular JSON object.
+             *
+             * Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create#custom_and_null_objects
+             */
+            const data = { ...originalData };
```
phieronymus, line 91: You may conserve the `null` or `undefined` values for the data. And, in a general manner, the type of the data if not an object.

laxa88: oh right... 🤔 i didn't think about that. thanks! let me update this.

laxa88: added explicit object type check, hopefully this is more specific 🙇

phieronymus (approved):
Thanks. Merging this.

PR #194, opened 2021-11-15
Update platform configuration

phieronymus:
No ticket linked to this PR, fixing the types to fix the platform creation in the tests.

### DIFF
--- a/src/resources/common/Platform.ts
+++ b/src/resources/common/Platform.ts
@@ -12,7 +12,20 @@ type CheckoutConfiguration = {
     ecProducts: { enabled: boolean };
 };
 
+export enum TransferMatchAmount {
+    MAXIMUM = "maximum",
+    MINIMUM = "minimum",
+    EXACT = "exact",
+}
+
+export type BankTransferConfiguration = {
+    enabled: boolean;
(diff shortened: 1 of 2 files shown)


src/resources/common/Platform.ts
```diff
@@ -12,7 +12,20 @@ type CheckoutConfiguration = {
...
+export enum TransferMatchAmount {
+    MAXIMUM = "maximum",
+    MINIMUM = "minimum",
+    EXACT = "exact",
+}
+
+export type BankTransferConfiguration = {
+    enabled: boolean;
+    matchAmount: TransferMatchAmount;
+    expiration: string;
```
laxa88, line 24: hmm... i'd call this `expireDate` to better describe the `string`, but not a big deal 😄

phieronymus: Let me double check the type here 🙂

phieronymus: Got the fixture wrong. It seems to be the expiration period. Somehow...

laxa88: oh nice! thanks for the indirect fix XD

laxa88 (approved):

PR #208, opened 2021-11-22
Add poll subscription and first charge helper

phieronymus:
Will be needed to make the code in the widget and checkout easier as well as for the service-typescript for the ec-email payment route.

### DIFF
--- a/src/resources/Subscriptions.ts
+++ b/src/resources/Subscriptions.ts
@@ -7,9 +7,11 @@ import { AuthParams, HTTPMethod, PollParams, ResponseCallback, RestAPI, SendData
 import { ProcessingMode } from "./common/enums";
 import { ignoreDescriptor } from "./common/ignoreDescriptor";
 import { Metadata, WithStoreMerchantName } from "./common/types";
+import { Charges } from "./Charges";
 import { ChargesListParams, ResponseCharges } from "./Charges";
 import { CRUDItemsResponse, CRUDPaginationParams, CRUDResource } from "./CRUDResource";
 import { PaymentType } from "./TransactionTokens";
+import { ChargeStatus, ResponseCharge } from ".";
 
 export enum SubscriptionPeriod {
     DAILY = "daily",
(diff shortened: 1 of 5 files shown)


src/resources/Subscriptions.ts
```diff
@@ -358,6 +367,37 @@ export class Subscriptions extends CRUDResource {
...
+    async pollSubscriptionWithFirstCharge(
+        storeId: string,
+        id: string,
+        data?: SendData<PollParams>,
+        auth?: AuthParams,
+        callback?: ResponseCallback<{ subscription: ResponseSubscription; charge?: ResponseCharge }>
+    ): Promise<{ subscription: ResponseSubscription; charge?: ResponseCharge }> {
+        const subscription = await this.poll(storeId, id, data);
+
+        const charge = await (async () => {
```
perrin4869, line 379:
```js
        const charge = !hasImmediateCharge(subscription)
            ? null
            : await this.api.longPolling(
                () => this.charges(storeId, id),
                (charges) => !!charges.items.length
            ).then(({ items: charges }) => this.api.longPolling(
                () => this.chargesResource.get(charges[0].storeId, charges[0].id, data, auth),
                ({ status }) => status !== ChargeStatus.PENDING
            ));
```

This is more compact than the other version and avoids the immediately called anonymous function, but both are ultimately equivalent!

perrin4869 (approved):
Looks good! Left just one suggestion

src/resources/Subscriptions.ts
phieronymus: Sure, let me refactor this part 🙂 Thanks!

phieronymus: Also used `charges.poll` instead of re-implementing the long polling for this resource. Thanks for pointing this function out 🙂

src/resources/Subscriptions.ts
```diff
@@ -256,18 +258,25 @@ export class ScheduledPayments extends CRUDResource {
     }
 }
 
+export const hasImmediateCharge = (subscription: SubscriptionItem) => {
```
laxa88, line 261: doesn't seem like you need to export this 🤔

phieronymus: Right. Over-engineering. I thought it would be useful if we want to do some custom implementation still but you are right, this is not needed, Let me remove the export 🙂

laxa88 (approved):

PR #258, opened 2022-01-14
Feature/laxa88/test

laxa88:
Refactor the blob test

### DIFF
--- a/test/specs/object.specs.ts
+++ b/test/specs/object.specs.ts
@@ -91,10 +91,11 @@ describe("Object Helpers", () => {
     describe("isBlob", () => {
         it("should return true when the parameter is a blob", function () {
             const version = process.version?.split(".");
-            const major = +version[0].substr(1); // e.g. 'v14' -> '14'
-            const minor = +version[1];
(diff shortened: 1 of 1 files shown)


test/specs/object.specs.ts
```diff
@@ -90,8 +90,13 @@ describe("Object Helpers", () => {
...
         it("should return true when the parameter is a blob", function () {
-            if (process.version.includes("v12")) {
-                // node 12 does not support blob
+            const version = process.version?.split(".");
+            const major = +version[0]?.substr(1) || 0; // e.g. 'v14' -> '14'
+            const minor = +version[1] || 0;
+            const currVersion = [major, minor].join(".");
+
+            if (currVersion < "14.18") {
+                // Blob was introduced in v14.18, so skip older versions
```
phieronymus, line 99: Sounds good 🙂 

Just curiosity. Is there a need for the `|| 0` and such? `process.version < "v14.18.0"` does not work properly?

laxa88: huh... i overcomplicated the code! thank you for pointing it out, let me fix it.

phieronymus (approved):

PR #365, opened 2022-04-18
Fix blob handling

phieronymus:
Fixes an issue when using the one touch PUT route

### DIFF
--- a/src/api/RestAPI.ts
+++ b/src/api/RestAPI.ts
@@ -22,7 +22,7 @@ import { RequestError } from "../errors/RequestResponseError";
 import { TimeoutError } from "../errors/TimeoutError";
 import { ProcessingMode } from "../resources/common/enums";
 import { checkStatus, parseJSON } from "../utils/fetch";
-import { toSnakeCase, transformKeys } from "../utils/object";
+import { isBlob, toSnakeCase, transformKeys } from "../utils/object";
 
 import { extractJWT, JWTPayload, parseJWT } from "./utils/JWT";
 import { containsBinaryData, objectToFormData } from "./utils/payload";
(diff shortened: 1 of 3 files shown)


src/api/RestAPI.ts
```diff
@@ -100,8 +100,10 @@ export type SendData<Data> = Data;
 
 const { stringify } = JSONBig({ useNativeBigInt: true });
 
-const getRequestBody = <Data>(data: SendData<Data>, keyFormatter = toSnakeCase): string | FormData =>
-    containsBinaryData(data)
+const getRequestBody = <Data>(data: SendData<Data>, keyFormatter = toSnakeCase): string | FormData | Blob =>
+    isBlob(data)
+        ? data
```
phieronymus, line 105: Do not try to transform the keys of the object when it is a blob

src/api/utils/payload.ts
```diff
@@ -50,7 +50,7 @@ export const objectToFormData = (obj: unknown, keyFormatter = toSnakeCase, ignor
         }
     };
 
-    appendFormData(transformKeys(obj, keyFormatter, ignoredKeys));
+    appendFormData(isBlob(obj) ? obj : transformKeys(obj, keyFormatter, ignoredKeys));
```
phieronymus, line 53: Same here. No need to transform the keys when it is a blob

src/resources/Resource.ts
```diff
@@ -58,12 +58,23 @@ type ObjectType =
...
     | "symbol"
-    | "undefined";
+    | "undefined"
+    | "blob";
 
-const getObjectType = (o: unknown): ObjectType => {
-    if (o === null) return "null";
-    if (Array.isArray(o)) return "array";
-    return typeof o;
+const getObjectType = (data: unknown): ObjectType => {
```
phieronymus, line 64: This function was added to destructure form data in order to prevent adding the multipart data for JSON. We just need to be sure we do not destructure the blobs as well to prevent the loss of the important class properties.

YusukeFukushima (approved):

PR #410, opened 2022-05-26
Switch to nodenext module

perrin4869:
Merging this will ensure we maintain compatibility with `nodenext` modules (for example, we don't forget to add `.js` extensions).
Unfortunately, this is currently blocked by https://github.com/auth0/jwt-decode/pull/130, and maybe even by `ts-node`, so keeping it as a draft

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+coverage
+dist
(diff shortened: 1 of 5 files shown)


perrin4869:
ok I managed to find a way to override the official typings temporarily at least
We can remove those if/when https://github.com/auth0/jwt-decode/pull/130 is ever merged

phieronymus (approved):
Thanks!

PR #413, opened 2022-05-30
Fix/pierreh/refactors

phieronymus:
A bit more refactoring. Thanks @perrin4869 for pointing the library to simplify the code out.

### DIFF
--- a/test/specs/refunds.specs.ts
+++ b/test/specs/refunds.specs.ts
@@ -4,10 +4,8 @@ import * as sinon from "sinon";
 import { SinonSandbox } from "sinon";
 import { v4 as uuid } from "uuid";
 
-import { HTTPMethod, RestAPI } from "../../src/api/RestAPI.js";
-import { POLLING_INTERVAL, POLLING_TIMEOUT } from "../../src/common/constants.js";
+import { RestAPI } from "../../src/api/RestAPI.js";
 import { RequestError } from "../../src/errors/RequestResponseError.js";
-import { TimeoutError } from "../../src/errors/TimeoutError.js";
 import {
     RefundCreateParams,
     RefundReason,
@@ -19,6 +17,7 @@ import { createRequestError } from "../fixtures/errors.js";
 import { generateList } from "../fixtures/list.js";
 import { generateFixture as generateRefund } from "../fixtures/refund.js";
 import { testEndpoint } from "../utils/index.js";
+import { assertPoll, assertPollCancel, assertPollTimeout } from "../utils/poll-helper.js";
 import { pathToRegexMatcher } from "../utils/routes.js";
 
 describe("Refunds", () => {
@@ -96,6 +95,10 @@ describe("Refunds", () => {
     });
 
     context("GET /stores/:storeId/charges/:chargeId/refunds/:id", () => {
+        const successItem = recordData;
+        const pendingItem = { ...recordData, status: RefundStatus.PENDING };
+        const failingItem = { ...recordData, status: RefundStatus.FAILED };
+
         it("should get response", async () => {
             fetchMock.getOnce(recordPathMatcher, {
                 status: 200,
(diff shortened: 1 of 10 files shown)


test/utils/poll-helper.ts
```diff
@@ -0,0 +1,131 @@
+import { expect } from "chai";
+import fetchMock from "fetch-mock";
+import * as sinon from "sinon";
+
+import { HTTPMethod } from "../../src/api/RestAPI.js";
+import { POLLING_INTERVAL, POLLING_TIMEOUT } from "../../src/common/constants.js";
+import { TimeoutError } from "../../src/errors/TimeoutError.js";
+
+const tickPoll = async (sandbox: sinon.SinonSandbox, ticks = 3) => {
```
perrin4869, line 9: `for await`, use async iterators here
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of

test/utils/poll-helper.ts
```diff
@@ -0,0 +1,131 @@
...
+    for (let i = 0; i < ticks; i++) {
+        await sandbox.clock.tickAsync(POLLING_INTERVAL);
+    }
+};
+
+const timeoutPoll = async (sandbox: sinon.SinonSandbox) => {
```
perrin4869, line 15: no need for `async/await` here, just return `sandbox.clock.tickAsync`

test/specs/refunds.specs.ts
```diff
@@ -107,99 +110,19 @@ describe("Refunds", () => {
...
-                }
-            );
-
-            const request = refunds.poll(uuid(), uuid(), uuid());
-            await sandbox.clock.tickAsync(POLLING_INTERVAL);
-            await sandbox.clock.tickAsync(POLLING_INTERVAL);
-
-            await expect(request).to.eventually.eql(recordData);
+            const call = () => refunds.poll(uuid(), uuid(), uuid());
+            await assertPoll(recordPathMatcher, call, sandbox, successItem, pendingItem);
```
perrin4869, line 114: no need for `async/await`, you can return the promise :)

perrin4869 (changes requested):
Looking clean!!!

test/specs/refunds.specs.ts
phieronymus: Most of the other tests are using async-await instead of returning the promise 🤔 I feel it would be more consistent in the univapay-node to keep it this way unless we want to engage in a giant test refactoring.

test/utils/poll-helper.ts
```diff
@@ -0,0 +1,127 @@
...
+import fetchMock from "fetch-mock";
+import pMap from "p-map";
+import * as sinon from "sinon";
+
+import { HTTPMethod } from "../../src/api/RestAPI.js";
+import { POLLING_INTERVAL, POLLING_TIMEOUT } from "../../src/common/constants.js";
+import { TimeoutError } from "../../src/errors/TimeoutError.js";
+
+const tickPoll = async (sandbox: sinon.SinonSandbox, ticks = 3) =>
+    pMap(Array(ticks).fill(0), async () => sandbox.clock.tickAsync(POLLING_INTERVAL), { concurrency: 1 });
```
perrin4869, line 11: no need for the `async` keyword

test/utils/poll-helper.ts
```diff
@@ -0,0 +1,127 @@
...
+
+const timeoutPoll = async (sandbox: sinon.SinonSandbox) => sandbox.clock.tickAsync(POLLING_TIMEOUT);
```
perrin4869, line 13: no need for the `async` keyword

test/utils/poll-helper.ts
```diff
@@ -0,0 +1,127 @@
...
+
+export const assertPoll = async <Item>(
+    pathMatcher: fetchMock.MockMatcher,
+    call: () => Promise<Item>,
+    sandbox: sinon.SinonSandbox,
+    successItem: Item,
+    pendingItem: Item | Item[]
+) => {
+    const pendingItems = Array.isArray(pendingItem) ? pendingItem : [pendingItem];
```
perrin4869, line 22: maybe use [`arrify`](https://www.npmjs.com/package/arrify) here?

test/specs/refunds.specs.ts
perrin4869: Sure, was just thinking that if the test is just one `await` at the end, use `return` rather than `await`, but since this isn't enforced by the CI it's fine either way!

perrin4869 (approved):
Few minor commens!

test/utils/poll-helper.ts
phieronymus: Sounds good to me. Thanks! 🙂

perrin4869 (approved):

PR #425, opened 2022-06-01
Add user-agent header

### LINKED ISSUE #421: Add User-Agent header to log univapay-node version
This would help logging the versions in use.

perrin4869:
Closes #421 

Mostly wanted to know what type of client people are using this sdk on.
I would have liked to add an `SDKVersion` info param in too, but I couldn't find a way to fetch it from the `package.json` file in such a way that it would work for all `cjs`, `esm`, `browser`, `node` environments, etc


kgston (commented):
Looks good. I think its a good idea.

src/api/RestAPI.ts
```diff
@@ -271,6 +272,8 @@ export class RestAPI extends EventEmitter {
             jwt = this.jwtRaw,
         } = auth || {};
 
+        headers.append("User-Agent", userAgent());
```
phieronymus, line 275: 👍

src/api/utils/userAgent.ts
```diff
@@ -0,0 +1,26 @@
...
+    if (isBrowser) {
+        const { browser: { name, version } } = bowser.parse(window.navigator.userAgent);
+
+        return `${name}@${version}`;
+    }
+
+    return "uknown-platform"
+}
+
+function getModuleType() {
```
phieronymus, line 20: Normally we try to go for arrow function when we do not need to bind. But the nodes being very weird feel free to ignore this coment.

src/api/utils/userAgent.ts
```diff
@@ -0,0 +1,26 @@
...
+    return typeof require === 'function' ? 'commonjs' : 'module';
+}
+
+export function userAgent() {
+    return  `${CLIENT_NAME} ${getPlatform()} ${getModuleType()}`;
```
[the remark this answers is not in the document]
perrin4869: Nice catch! `eslint-config-prettier` was not setup correctly

src/api/RestAPI.ts
```diff
@@ -199,7 +200,7 @@ export class RestAPI extends EventEmitter {
             });
         }
 
-        const payload: boolean = [HTTPMethod.GET, HTTPMethod.DELETE].indexOf(method) === -1;
+        const payload: boolean = [HTTPMethod.GET, HTTPMethod.DELETE].includes(method) === false;
```
perrin4869, line 203: @phieronymus just a warning that `IE11` doesn't support `Array.prototype.includes`, so it needs to be polyfilled if it's not already

phieronymus:
@perrin4869 Thanks a lot for the fixes 🙂 

Note that we have the polyfills on the consoles and I would expect the user to have them as well if they desire to support IE11.

We already have a function using it:

```tsx
const isSymbol = (value: unknown): value is string | number | boolean =>
    ["string", "number", "boolean"].includes(typeof value);
```

phieronymus (approved):

PR #433, opened 2022-06-02
p-timeout v4

perrin4869:
v5 is esm, but v4 should still be fine


phieronymus:
Sure! Thanks for catching this 🙂

phieronymus (approved):

PR #436, opened 2022-06-03
Remove David DM badges from readme

perrin4869:
Those badges are no longer working for a long time now


phieronymus (approved):
Oh ok. Thanks for checking! I did not look at the badges for so long that I actually forgot they existed.

PR #437, opened 2022-06-03
Remove hound

perrin4869:
Doesn't seem to be used anymore?


phieronymus (approved):
Even though it feels always weird to remove those features I must agree that hound was kind of useless for a very long time. Let's get rid of the complexity :+1:

PR #450, opened 2022-06-10
Allow api.send to override endpoint

perrin4869:
I noticed a hack in the `oneTouch` API for overriding the endpoint.
Trying to formalize this here, will send a PR to the private SDK shortly with the second half.
This should be released as `2.1`


phieronymus (approved):
Sounds good to me. Thanks for cleaning this up.

PR #457, opened 2022-06-16
Reove mutation and context bind in defineRoute

phieronymus:
As I was investigating something I noticed that the defineRoute function got more complex than it should. Decided to refactor it.

### DIFF
--- a/src/resources/Resource.ts
+++ b/src/resources/Resource.ts
@@ -102,16 +102,14 @@ export abstract class Resource extends EventEmitter {
         acceptType?: string,
         keyFormatter = toSnakeCase
     ): DefinedRoute {
-        const api: RestAPI = this.api;
-
-        return function route<A, B>(
+        return <A, B>(
             originalData?: SendData<A>,
             callback?: ResponseCallback<B>,
             auth?: AuthParams,
             pathParams: Record<string, string> = {}
-        ): Promise<B> {
+        ): Promise<B> => {
             /**
-             * Sanitises and ensures that the data is recreated as an object with default prototypes.
+             * Sanitizes and ensures that the data is recreated as an object with default prototypes.
              * In rare cases if the data was created with `Object.create(null)` (no prototypes),
              * this will cause the data to be sent as `WebKitFormBoundary` instead of regular JSON object.
              *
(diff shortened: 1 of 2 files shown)


phieronymus:
Seems that I broke something. I'll check soon

phieronymus:
Somehow found the issue and it hurt my ego to have been missing this...

src/resources/Resource.ts
```diff
@@ -102,43 +102,48 @@ export abstract class Resource extends EventEmitter {
...
              *
              * Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create#custom_and_null_objects
              */
             const data = getObjectType(originalData) === "object" ? { ...originalData } : originalData;
 
-            const url: string = compilePath(path, pathParams);
+            const url = compilePath(path, pathParams);
 
-            const missingPathParams: string[] = (url.match(/:([a-z]+)/gi) || []).map((m: string) => m.replace(":", ""));
-            const missingParams: string[] = missingKeys(data, required);
```
perrin4869, line 125: This could be made more efficient by doing `obj ? required.find(key => data[key] === undefined) : keys[0]`.
Since we are only interested in the first missing param

src/resources/Resource.ts
```diff
@@ -102,43 +102,48 @@ export abstract class Resource extends EventEmitter {
...
              * this will cause the data to be sent as `WebKitFormBoundary` instead of regular JSON object.
              *
              * Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create#custom_and_null_objects
              */
             const data = getObjectType(originalData) === "object" ? { ...originalData } : originalData;
 
-            const url: string = compilePath(path, pathParams);
+            const url = compilePath(path, pathParams);
 
-            const missingPathParams: string[] = (url.match(/:([a-z]+)/gi) || []).map((m: string) => m.replace(":", ""));
```
perrin4869, line 124: Use `Array.prototype.find` here too?

src/resources/Resource.ts
```diff
@@ -102,43 +102,48 @@ export abstract class Resource extends EventEmitter {
...
+            return this.api.send(
+                method,
+                url,
+                data,
+                auth,
+                callback,
+                requireAuth,
+                acceptType,
+                keyFormatter
+            ) as Promise<B>;
```
perrin4869, line 146: Maybe `api.send` should be polymorphic instead of casting?

perrin4869 (approved):

src/resources/Resource.ts
phieronymus: Nice catch! Actually the map is here to format. Let me only format the first parameter in this case 🙂

src/resources/Resource.ts
phieronymus: Yeah! I could remove the other function, nice catch!

src/resources/Resource.ts
```diff
@@ -102,43 +102,48 @@ export abstract class Resource extends EventEmitter {
...
-            let err: Error;
-
-            if (missingPathParams.length > 0) {
-                err = fromError(new PathParameterError(missingPathParams[0]));
-                callback?.(err);
-                return Promise.reject(err);
+            const url = compilePath(path, pathParams);
+
+            // Validate required path parameters
+            const firstMissingPathParam = (url.match(/:([a-z]+)/gi) || [])[0]?.replace(":", "");
```
perrin4869, line 122: Can further simply `url.match(/:([a-z]+)/gi)?.[0]?.replace(":", "");`

PR #467, opened 2022-06-21
Add handling 500 error

YusukeFukushima:
Fixes https://github.com/univapaycast/gopay-checkout/issues/656

### DIFF
--- a/src/api/RestAPI.ts
+++ b/src/api/RestAPI.ts
@@ -20,7 +20,7 @@ import {
 } from "../common/constants.js";
 import { RequestErrorCode, ResponseErrorCode } from "../errors/APIError.js";
 import { fromError } from "../errors/parser.js";
-import { RequestError } from "../errors/RequestResponseError.js";
+import { RequestError, ResponseError } from "../errors/RequestResponseError.js";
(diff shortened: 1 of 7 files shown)


src/api/RestAPI.ts
```diff
@@ -336,25 +337,37 @@ export class RestAPI extends EventEmitter {
...
-
-                if (cancelCondition?.(result)) {
-                    return null;
-                }
-
-                if (!successCondition(result)) {
-                    await new Promise((resolve) => setTimeout(resolve, interval)); // sleep to avoid firing requests too fast
-                    return repeater();
+        let internalErrrorCount = 0;
+        while (internalErrrorCount <= MAX_INTERNAL_ERROR_RETRY) {
```
phieronymus, line 341: You likely do not need the `while` here if you move the try/catch block around `const result = await promise();` only which would make the logic easier to read as you wont need to follow the returns in the loop.

YusukeFukushima:
@phieronymus 
Thanks a lot for your fixing :bow: 
I think there is no problem in this PR now.

phieronymus (approved):

PR #468, opened 2022-06-22
Open poll options for long polling

phieronymus:
We will need to set the interval in some occasion and there is not reason to now open those options to dev and users.


perrin4869 (approved):
The options for many of these functions are getting crowded with optional arguments.
Maybe for next major release we can cleanup the arguments a little bit?

phieronymus:
@perrin4869 Opened https://github.com/univapay/univapay-node/issues/469 for the cleanup. This PR will not handle a big refactoring.

perrin4869:
Of course! That would be a breaking change too, maybe next year when we drop node 14 haha

PR #471, opened 2022-06-23
Group poll params

phieronymus:
Updates https://github.com/univapay/univapay-node/issues/469

Some refactoring for the poll arguments 🙂

### DIFF
--- a/src/api/RestAPI.ts
+++ b/src/api/RestAPI.ts
@@ -30,6 +30,33 @@ import { extractJWT, JWTPayload, parseJWT } from "./utils/JWT.js";
 import { containsBinaryData, objectToFormData } from "./utils/payload.js";
 import { userAgent } from "./utils/userAgent.js";
 
+export type PollParams<Response> = {
+    /**
+     * Condition for which the response is considered to be successful and stop the polling
+     */
+    successCondition: (response: Response) => boolean;
+
+    /**
+     * Condition to cancel the polling without triggering error
+     */
+    cancelCondition?: (response: Response) => boolean;
+
+    /**
+     * Time after which a TimeoutError will be triggered and the poll will be canceled
+     */
+    timeout?: number;
+
+    /**
+     * Interval between polls
+     */
+    interval?: number;
+
+    /**
+     * Callback to be triggered at each poll iteration
+     */
+    iterationCallback?: (response: Response) => void;
+};
+
 export enum HTTPMethod {
     GET = "GET",
     POST = "POST",
@@ -94,9 +121,9 @@ export interface AuthParams {
     appId?: string;
 }
 
-export interface PollParams {
+export type PollData = {
     polling?: boolean;
-}
+};
 
 export type PromiseCreator<A> = () => Promise<A>;
 
@@ -302,43 +329,22 @@ export class RestAPI extends EventEmitter {
      * @internal
      */
     async longPolling<Response>(
-        /**
-         * API call to be repeated
-         */
         promise: PromiseCreator<Response>,
-
-        /**
-         * Condition for which the response is considered to be successful and stop the polling
-         */
-        successCondition: (response: Response) => boolean,
-
-        /**
-         * Condition to cancel the polling without triggering error
-         */
-        cancelCondition?: (response: Response) => boolean,
-
-        /**
-         * Callback to be triggered at the end of the long polling
-         */
-        callback?: ResponseCallback<Response>,
-
-        /**
-         * Time after which a TimeoutError will be triggered and the poll will be canceled
-         */
-        timeout: number = POLLING_TIMEOUT,
-
-        /**
-         * Interval between polls
-         */
-        interval: number = POLLING_INTERVAL,
-
-        /**
-         * Callback to be triggered at each poll iteration
-         */
-        iterationCallback?: (response: Response) => void
+        pollParams: PollParams<Response>,
+        callback?: ResponseCallback<Response>
     ): Promise<Response> {
+        const {
+            successCondition,
+            cancelCondition,
+            iterationCallback,
+            timeout = POLLING_TIMEOUT,
+            interval = POLLING_INTERVAL,
+        } = pollParams;
+
         return execRequest(async () => {
             let internalErrorCount = 0;
+
+            const sleepInterval = () => new Promise((resolve) => setTimeout(resolve, interval));
 
             const repeater = async (): Promise<Response> => {
                 try {
@@ -351,17 +357,19 @@ export class RestAPI extends EventEmitter {
                     }
 
                     if (!successCondition(result)) {
-                        await new Promise((resolve) => setTimeout(resolve, interval)); // sleep to avoid firing requests too fast
+                        sleepInterval();
                         return repeater();
                     }
 
                     return result;
                 } catch (error) {
+                    // Use retry mechanism for 500 as internal server error on API side do not always mean failure
                     if (error.errorResponse?.httpCode !== 500 || internalErrorCount >= MAX_INTERNAL_ERROR_RETRY) {
                         throw error;
                     }
 
                     internalErrorCount++;
+                    sleepInterval();
                     return repeater();
                 }
             };
--- a/src/api/RestAPI.ts
+++ b/src/api/RestAPI.ts
@@ -105,11 +105,6 @@ export interface ErrorResponse {
     errors: (SubError | ValidationError)[];
 }
 
-export type ResponseCallback<A> = (response: A | Error) => void;
-
-export type PromiseResolve<A> = (value?: A | PromiseLike<A>) => void;
-export type PromiseReject = (reason?: any) => void;
-
 export interface AuthParams {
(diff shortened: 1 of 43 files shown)


src/api/RestAPI.ts
```diff
@@ -351,17 +357,19 @@ export class RestAPI extends EventEmitter {
                     }
 
                     if (!successCondition(result)) {
-                        await new Promise((resolve) => setTimeout(resolve, interval)); // sleep to avoid firing requests too fast
+                        sleepInterval();
```
perrin4869, line 360: `await` ;p

src/api/RestAPI.ts
```diff
@@ -351,17 +357,19 @@ export class RestAPI extends EventEmitter {
...
 
                     return result;
                 } catch (error) {
+                    // Use retry mechanism for 500 as internal server error on API side do not always mean failure
                     if (error.errorResponse?.httpCode !== 500 || internalErrorCount >= MAX_INTERNAL_ERROR_RETRY) {
                         throw error;
                     }
 
                     internalErrorCount++;
+                    sleepInterval();
```
perrin4869, line 372: `await`

src/resources/Charges.ts
```diff
@@ -130,29 +130,15 @@ export class Charges extends CRUDResource {
...
         callback?: ResponseCallback<ResponseCharge>,
-
-        /**
-         * Condition for the resource to be successfully loaded. Default to pending status check.
-         */
-        cancelCondition?: (response: ResponseCharge) => boolean,
-        successCondition: ({ status }: ResponseCharge) => boolean = ({ status }) => status !== ChargeStatus.PENDING,
-        iterationCallback?: (response: ResponseCharge) => void,
-        pollOptions?: { interval?: number; timeout?: number }
+        pollParams?: Partial<PollParams<ResponseCharge>>
```
perrin4869, line 136: hm... `pollParams` coming the `callback` is somewhat iffy... I'm almost tempted to say, let's get rid of the callbacks...

perrin4869 (commented):
Change looks awesome, but it's kinda breaking ^^;;
Did we gather some data on who is using the SDK?

src/api/RestAPI.ts
phieronymus: Me no smart. Thanks a lot!

src/resources/Charges.ts
phieronymus: I tried to keep parameters common to all requests together.

I am also tempted to remove the callback and let people use `.then` to be honest...

phieronymus: Or maybe should we use the same for `auth` and `callback`? Like `requestParams`? 🤔

phieronymus:
@perrin4869 Yeah... This is a breaking change. I actually think we will have to go through that to limit future breaking changes as features and options are growing though 🤔 If we do a breaking release we may have the old user on the older version and we do not really have any bugs right now so it feels like it is the right timing.

phieronymus:
Very stupids mistakes. Sorry for that. The PR should be fixed now 🙂

src/resources/Cancels.ts
```diff
@@ -80,29 +80,15 @@ export class Cancels extends CRUDResource {
...
-        successCondition: ({ status }: ResponseCancel) => boolean = ({ status }) => status !== CancelStatus.PENDING,
-        iterationCallback?: (response: ResponseCancel) => void,
-        pollOptions?: { interval?: number; timeout?: number }
+        pollParams?: Partial<PollParams<ResponseCancel>>
     ): Promise<ResponseCancel> {
-        const pollingData = { ...data, polling: true };
-        const promise: () => Promise<ResponseCancel> = () => this.get(storeId, chargeId, id, pollingData, auth);
+        const pollData = { ...data, polling: true };
+        const promise: () => Promise<ResponseCancel> = () => this.get(storeId, chargeId, id, pollData, auth);
+        const successCondition = pollParams?.successCondition || (({ status }) => status !== CancelStatus.PENDING);
```
perrin4869, line 90: `||` -> `??`

src/resources/Charges.ts
perrin4869: If we make a breaking release I think remove all callbacks from the SDKs...

src/resources/Subscriptions.ts
```diff
@@ -359,48 +359,33 @@ export class Subscriptions extends CRUDResource {
...
-        successCondition: ({ status }: ResponseSubscription) => boolean = ({ status }) =>
-            status !== SubscriptionStatus.UNVERIFIED,
-        iterationCallback?: (response: ResponseSubscription) => void,
-        pollOptions?: { interval?: number; timeout?: number }
+        pollParams?: Partial<PollParams<ResponseSubscription>>
     ): Promise<ResponseSubscription> {
         const pollData = { ...data, polling: true };
         const promise: () => Promise<ResponseSubscription> = () => this.get(storeId, id, pollData, auth);
+        const successCondition =
+            pollParams?.successCondition || (({ status }) => status !== SubscriptionStatus.UNVERIFIED);
```
perrin4869, line 370: `||` -> `??`

perrin4869 (changes requested):
We just made a major release 2 weeks ago, wish we'd shipped this in right then and there lol
Sure we can do `3.0`. I was hoping next time to go ahead and drop support for `commonjs` too, but might be too early?

phieronymus:
@perrin4869 It quickly changed into a fairly big PR. Please feel free to have a look and comment again though.

src/api/RestAPI.ts
```diff
@@ -186,11 +208,9 @@ export class RestAPI extends EventEmitter {
         uri: string,
         data?: SendData<Data>,
         auth?: AuthParams,
-        callback?: ResponseCallback<ResponseBody>,
-        requireAuth = true,
-        acceptType?: string,
-        keyFormatter = toSnakeCase
+        options?: ApiSendOptions
     ): Promise<ResponseBody | string | Blob | FormData> {
+        const { requireAuth = true, acceptType, keyFormatter = toSnakeCase } = options || {};
```
perrin4869, line 213: `||` -> `??` 😅
or maybe move this up to be a default parameter value? `options?: ApiSendOptions = {}`

src/api/RestAPI.ts
```diff
@@ -302,44 +322,22 @@ export class RestAPI extends EventEmitter {
...
+        } = pollParams;
+
         return execRequest(async () => {
             let internalErrorCount = 0;
 
+            const sleepInterval = () => new Promise((resolve) => setTimeout(resolve, interval));
+
             const repeater = async (): Promise<Response> => {
                 try {
                     const result = await promise();
```
perrin4869, line 343: maybe `promise` could be renamed as `task`? `PromiseCreator` -> `Task`?

src/api/utils/payload.ts
```diff
@@ -16,9 +16,12 @@ export const containsBinaryData = (data: unknown): boolean => {
     } else if (isClassInstance(data)) {
         return true;
     } else if (Array.isArray(data)) {
-        return data.reduce((result: boolean, value: any) => result || containsBinaryData(value), false);
+        return data.reduce((result: boolean, value: unknown) => result || containsBinaryData(value), false);
     } else if (isObject(data)) {
-        return Object.keys(data).reduce((result: boolean, key: any) => result || containsBinaryData(data[key]), false);
+        return Object.keys(data).reduce(
```
perrin4869, line 21: `Object.keys(data).some(...)`?

src/resources/BankAccounts.ts
```diff
@@ -71,49 +71,27 @@ export class BankAccounts extends CRUDResource {
...
     static routeBase = "/bank_accounts";
 
-    list(
-        data?: SendData<BankAccountsListParams>,
-        auth?: AuthParams,
-        callback?: ResponseCallback<ResponseBankAccounts>
-    ): Promise<ResponseBankAccounts> {
-        return this._listRoute()(data, callback, auth);
+    list(data?: SendData<BankAccountsListParams>, auth?: AuthParams): Promise<ResponseBankAccounts> {
+        return this._listRoute()(data, auth);
```
perrin4869, line 75: maybe `this._listRoute()` doesn't need to be called each time? This could be a private field `this.#list = this._listRoute()`?
Maybe define a `getter` to make it lazy?

perrin4869 (commented):
Some comments for now;

src/api/RestAPI.ts
phieronymus: Seems that other part of the application call it `executor`. Let me rename it this way for consistency. Thanks for catching this.

src/resources/BankAccounts.ts
phieronymus: Added a lazy load mechanic. Typescript does not really like the lazy getter thing 🤔 Or maybe I missed something?

phieronymus:
Sure, let me change the names then 🙂

phieronymus:
Merging this as well!

PR #484, opened 2022-06-30
Fix issue with testcafe hammerhead

phieronymus:
Actually easier to handle it here. I'll continue investigating why hammerhead does not like it though...


perrin4869 (approved):
Let's just add a comment with a link to the `testcafe hammerhead` issue?

phieronymus:
Here is the issue: https://github.com/DevExpress/testcafe/issues/6863

Following this thread:
https://github.com/DevExpress/testcafe/issues/6863

PR #500, opened 2022-07-11
Formalize languages

perrin4869:
I am not sure about the types here and how they relate to the front-end...
The services will accept `en_us`, `ja_jp` and `zh_tw` though


phieronymus (approved):
Sounds good to me.

Note that the keys will be snakify which is why it was camelized I guess.

PR #522, opened 2022-07-27
Add link product route on public sdk

phieronymus:
Open the link product route on the private SDK for checkout usage 🙂 

@perrin4869 Would it be possible to ensure that the route would work without a storeId and merchantId in the pathname, using the store application token as auth? Sorry that I missed that check during the previous review.

### DIFF
--- a/src/resources/LinksProducts.ts
+++ b/src/resources/LinksProducts.ts
@@ -9,7 +9,6 @@ import { TransactionTokenType } from "./TransactionTokens.js";
 
 export type LinkProductListItem = {
     id: string;
-    platformId: string;
     merchantId: string;
     storeId: string;
     name: string;
--- a/src/resources/LinksProducts.ts
+++ b/src/resources/LinksProducts.ts
@@ -1,0 +1,38 @@
+/**
+ *  @module Resources/Products
+ */
+
+import { AuthParams, SendData } from "../api/RestAPI.js";
+import { CRUDItemsResponse, CRUDResource } from "./CRUDResource.js";
+import { DefinedRoute } from "./Resource.js";
+import { TransactionTokenType } from "./TransactionTokens.js";
+
+export type LinkProductListItem = {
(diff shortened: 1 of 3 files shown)


src/resources/LinksProducts.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ *  @module Resources/Products
+ */
+
+import { AuthParams, SendData } from "../api/RestAPI.js";
+import { CRUDItemsResponse, CRUDResource } from "./CRUDResource.js";
+import { DefinedRoute } from "./Resource.js";
+import { TransactionTokenType } from "./TransactionTokens.js";
+
+export type LinkProductListItem = {
```
phieronymus, line 10: Also I did not include the active parameter in the public ones. I may want to remove the platformId as well, right?

perrin4869: Yeah `platformId` and `active` are omitted for non admins: https://github.com/univapaycast/services-typescript/blob/ea988d2a1f516fad8ebb496439edb550dc39ee47/services/checkout/src/handlers/links_products/list.ts#L34

src/resources/LinksProducts.ts
```diff
@@ -0,0 +1,38 @@
...
+    quantity: number;
+};
+export type ResponseLinkProducts = CRUDItemsResponse<LinkProductListItem> & {
+    hasMore: false; // no support for pagination
+};
+
+export class LinksProducts extends CRUDResource {
+    static routeBase = "/checkout/links/:linkId/products";
+
+    private _list?: DefinedRoute;
```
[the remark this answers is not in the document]
perrin4869: That is such a bother lol
Typescript be trippin'

perrin4869 (changes requested):
Just gotta remove the `platformId` I think!

perrin4869 (approved):

PR #525, opened 2022-07-29
Fix ec language types

phieronymus:
Used in the following type:

```
customFieldsTitles?: Partial<Record<Languages, string>> | null;
orderSummaryTitles?: Partial<Record<Languages, string>> | null;
customFields?: Partial<Record<Languages, ECFormCustomField[]>> | null;
```

The keys for those items are camelized on the SDK and snakified when formatting for the API. When using them the proper type is the camelized version as we want to use the front-end-formatted version. Please feel free to tell me if I overlook anything.

### DIFF
--- a/src/api/RestAPI.ts
+++ b/src/api/RestAPI.ts
@@ -139,14 +139,23 @@ export type ApiSendOptions = {
      * Custom formatter to format the request object to the API (skipped for blobs)
      */
     keyFormatter?: (key: string) => string;
+
+    /**
+     * List of key to ignore formatting on when parsing the response
+     */
+    ignoreKeysFormatting?: string[];
 };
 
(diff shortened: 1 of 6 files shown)


perrin4869:
Well... I'm using the values of these guys in [here](https://github.com/univapaycast/services-typescript/blob/af6500df9bea7ea82012cd55a2398bc6e1e789bb/services/checkout/src/schemas/common.ts#L52)
This is a case where I'm not sure camelizing is the way to go? I'd vote to keep them always snake case...

phieronymus:
The current type is wrong though 🤔 I can ignore those properties in the camelization process but it would make the code more complexe and be breaking. Should we get that as a quick fix and maybe change that for next time we want a breaking version?

phieronymus:
The thing is that we do not have much gain with having this inconsistency 🤔

perrin4869 (approved):
Nice!!! Thanks for this!!!

PR #570, opened 2022-09-12
Add remindNotificationEnabled

YusukeFukushima:
Updated https://github.com/univapaycast/univapay-admin-console/issues/1503

### DIFF
--- a/src/resources/common/Configuration.ts
+++ b/src/resources/common/Configuration.ts
@@ -101,7 +101,7 @@ export type BankTransferConfiguration = PaymentTypeConfiguration & {
     /**
      * Boolean when a reminder email should be sent
      */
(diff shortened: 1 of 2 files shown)


phieronymus (approved):

src/resources/common/Configuration.ts
```diff
@@ -98,6 +98,11 @@ export type BankTransferConfiguration = PaymentTypeConfiguration & {
      */
     automaticExtensionEnabled: boolean;
 
+    /**
+     * Boolean when a reminder email should be sent
+     */
+    remindNotificationEnabled?: boolean;
```
phieronymus, line 104: Oh, this is required, not optional, can only that be changed?

PR #594, opened 2022-10-05
chore: run dependabot weekly

perrin4869:
Is `query-string` still being ignored?


phieronymus (approved):
Nice! That will shut down a lot of spaming!

PR #628, opened 2022-11-15
Attempt to fix login issue and cleanup

phieronymus:
Mostly cleanups but also setting the jwtRaw before decoding it to ensure the call in parallel can be done. Not optimal but I think the console can actually fail the login because of it.

### DIFF
--- a/src/api/RestAPI.ts
+++ b/src/api/RestAPI.ts
@@ -71,7 +71,7 @@ export enum HTTPMethod {
 export interface RestAPIOptions {
     endpoint?: string;
     jwt?: string;
-    handleUpdateJWT?(jwt: string): void;
+    handleUpdateJWT?: (jwt: string) => void;
     secret?: string;
     origin?: string;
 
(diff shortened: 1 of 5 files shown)


src/api/RestAPI.ts
```diff
@@ -196,27 +197,24 @@ export class RestAPI extends EventEmitter {
...
 
         this.appId = options.appId || process.env[ENV_KEY_APP_ID];
         this.secret = options.secret || process.env[ENV_KEY_SECRET];
         this.authToken = options.authToken;
     }
 
     set jwtRaw(jwtRaw: string) {
-        this.jwt = parseJWT(jwtRaw);
         this._jwtRaw = jwtRaw;
+        this.jwt = parseJWT(jwtRaw);
```
phieronymus, line 217: This is the change... Although this is supposed to be synchronous after thinking about it... Really not sure what went wrong...

YusukeFukushima (approved):

PR #629, opened 2022-11-17
Remove unsafe expiry check

phieronymus:
User system date was affecting the auth logic and would cause a token expired error even though the token is valid


YusukeFukushima (approved):
Thanks!

phieronymus:
Discussed this with @jdebono on slack. Merging this. I'll release it afterward.

PR #667, opened 2022-12-23
Add hide cvv to ec links

phieronymus:
Updates https://github.com/univapaycast/univapay-form-checkout/issues/422


gpapako (approved):
Do we need any other changes with this? Like how is it going to be used?

gpapako:
Merging!

PR #733, opened 2023-03-14
Remove cyrex pay from QR brands

phieronymus:
Fixes https://univapay-systems.slack.com/archives/C0172GCPT4M/p1678703575213189?thread_ts=1678703575.213189&cid=C0172GCPT4M

Cyrex pay is a card gateway and not a QR gateway. Removing it from the QR gateways will make it appear in the correct place.

Linked to https://github.com/univapaycast/univapay-common/pull/1051


gpapako (approved):
LGTM! But should we add it to some other enum? Or is it already covered? I.e. Private label?

gpapako:
Merging!

PR #735, opened 2023-03-15
Fix refund threshold

phieronymus:


gdp88 (approved):

gdp88:
alright let's merge this for now and if more changes are needed (as per https://github.com/univapaycast/univapay-admin-console/issues/1656) we do this on a separate PR

PR #736, opened 2023-03-15
Remove unused param

phieronymus:


gdp88 (approved):

gdp88:
alright and with that we should be good! merging!

PR #737, opened 2023-03-15
refactor: simplify api

perrin4869:
just noticed this while looking at https://github.com/univapaycast/univapay-node-private/pull/863


phieronymus:
Thanks!

PR #744, opened 2023-03-22
Add extra parameter in TransactioTokenCardDetails

gpapako:
Fixes: https://github.com/univapaycast/univapay-admin-console/issues/1665


phieronymus (approved):
Perfect! Thanks!

PR #756, opened 2023-04-06
Add fetchErrorReasons

YusukeFukushima:
Related to https://github.com/univapaycast/univapay-common/pull/1057

### DIFF
--- a/src/errors/RequestResponseError.ts
+++ b/src/errors/RequestResponseError.ts
@@ -8,7 +8,7 @@ import { ErrorResponse, SubError, ValidationError } from "../api/RestAPI.js";
 import { RequestErrorCode, ResponseErrorCode } from "./APIError.js";
 
 export type ErrorItem = boolean | number | string | SubError | ValidationError;
-type RawErrorRequest = {
+export type RawErrorRequest = {
     httpCode?: number;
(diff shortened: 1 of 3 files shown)


src/errors/RequestResponseError.ts
```diff
@@ -64,6 +64,16 @@ const serializeErrorResponse = ({ code, httpCode, errors }: FormattedErrorReques
...
 
+export const fetchErrorReasons = ({ errors, code }: FormattedErrorRequest) => {
+    const formattedErrors = arrify(errors);
+
+    if (formattedErrors.length === 0) {
+        return code;
+    }
+
+    return formattedErrors.map((error) => error.reason);
+};
```
phieronymus, line 75: Thanks. We have some error parsing function in the node 🙂 We may want to add this extra logic to the error parsing and add the unit tests.

Thanks a lot for adding that! 🎉 It will be super useful on the SDK

YusukeFukushima: @phieronymus 
Thanks for checking :bow: 
But, could you tell me more about the following?
> We may want to add this extra logic to the error parsing

(About the test, I'll add it now :+1: )

phieronymus (approved):

PR #757, opened 2023-04-07
Fix fetchErrorReasons

YusukeFukushima:
Related to https://github.com/univapaycast/univapay-common/pull/1057

@phieronymus 
Sorry, for repeatation. I made a mistake on the type of args.
Please check again :bow:

### DIFF
--- a/src/errors/RequestResponseError.ts
+++ b/src/errors/RequestResponseError.ts
@@ -73,7 +73,7 @@ export const fetchErrorReasons = (errorRequest: ErrorResponse) => {
         return code;
     }
 
-    return arrifiedErrors.map((error) => error.reason);
+    return arrifiedErrors;
 };
 
 export class RequestResponseBaseError extends Error {
--- a/src/errors/RequestResponseError.ts
+++ b/src/errors/RequestResponseError.ts
@@ -8,7 +8,7 @@ import { ErrorResponse, SubError, ValidationError } from "../api/RestAPI.js";
 import { RequestErrorCode, ResponseErrorCode } from "./APIError.js";
 
 export type ErrorItem = boolean | number | string | SubError | ValidationError;
-export type RawErrorRequest = {
+type RawErrorRequest = {
     httpCode?: number;
     code: ResponseErrorCode | RequestErrorCode;
     errors: ErrorItem | ErrorItem[];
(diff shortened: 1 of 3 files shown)


src/errors/RequestResponseError.ts
```diff
@@ -64,7 +64,7 @@ const serializeErrorResponse = ({ code, httpCode, errors }: FormattedErrorReques
     return `Code: ${code}, HttpCode: ${httpCode}, Errors: ${formattedErrors.join(", ")}`;
 };
 
-export const fetchErrorReasons = (errorRequest: RawErrorRequest) => {
+export const fetchErrorReasons = (errorRequest: ErrorResponse) => {
```
YusukeFukushima, line 67: @phieronymus 
Just to confirm, does the api ever return more than one error? If not, I think there is no need to return the reason as Array.

phieronymus: It is always one error response that can contain several errors in the `errors` array 🙂

YusukeFukushima: Okay. What I think is the error what is linked to the field is only one and we may want to specify the error. Which is good to add the logic, the SDK or common?

phieronymus: Sorry, I did not understand the previous message. Do you have an example?

YusukeFukushima: Sure. I am concerned about the following case.
```
{
    code: "VALIDATION_ERROR",
    status: "error",
    errors: [
        { field: "first-field", reason: "NUMBER_MIN" },
        { field: "second-field", reason: "CANNOT_BE_ZERO" },
    ]
}
```
On the current implementation, when `fetchErrorReasons` receive the above values, it returns `["NUMBER_MIN", "CANNOT_BE_ZERO"]`
But if it return these values, I am thinking that it is going to be hard to tell which error is from which field.
If the error object has always only one pair of field and reason, there are no problem, but actually it can have several pairs, so I think we need a logic to specify them.
In this case I also think we need to the way how to link the field.

Sorry if this is outlandish.

phieronymus: We will need to keep this format though (the pair with the field). You can then link them to the fields using the `mapAPIReponseToError` in the `Details` and `Create` components.

YusukeFukushima: Okay! Thanks :bow:

phieronymus (approved):

PR #765, opened 2023-04-17
Add subscription plans

phieronymus:
This does the following:

- Add subscription plans
- [Breaking] Remove fixed_cycles_amount from installment plans
- [Breaking] Fix not working type for installment plan


gpapako (approved):
LGTM! 👍

PR #772, opened 2023-04-20
Add subscription plan for configuration

phieronymus:
The same configuration has to be added to the private node

### DIFF
--- a/src/resources/common/Configuration.ts
+++ b/src/resources/common/Configuration.ts
@@ -169,7 +169,7 @@ export interface InstallmentsConfiguration {
 
 export interface InstallmentsConfigurationItem extends PaymentTypeConfiguration, InstallmentsConfiguration {}
 
-export type SubscriptionsPlanConfiguration = {
+export type SubscriptionPlanConfiguration = {
     enabled: boolean | null;
     fixedCycle: boolean | null;
     fixedCycleAmount: boolean | null;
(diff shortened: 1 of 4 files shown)


src/resources/common/Configuration.ts
```diff
@@ -169,6 +169,23 @@ export interface InstallmentsConfiguration {
 
 export interface InstallmentsConfigurationItem extends PaymentTypeConfiguration, InstallmentsConfiguration {}
 
+export type SubscriptionsPlanConfiguration = {
```
gdp88, line 172: very minor but if a variable name consists of multiple words I m more used to having them in their singular form and then use plural (if needed) on the last one. So here I would go with `SubscriptionPlanConfiguration`.
But seems like we have the same naming pattern for other `interface` names in this file as well so if we want to keep it consistent maybe we don't need to change 🙇

gdp88 (approved):

gdp88:
very minor comment..lgtm!

src/resources/common/Configuration.ts
phieronymus: Ups, thanks! That is correct, the type should follow the property name. Thanks for catching it.

gdp88: thanx! resolving!

gdp88:
alright merging this one!

PR #773, opened 2023-04-21
Add cyclical period to subscription

phieronymus:
Linked to https://univapay-systems.slack.com/archives/CA0U3CSQ1/p1681975187942809

### DIFF
--- a/src/resources/Subscriptions.ts
+++ b/src/resources/Subscriptions.ts
@@ -74,19 +74,18 @@ export type SubscriptionPlanItem =
     | { planType: InstallmentPlan.FIXED_CYCLES; fixedCycles: number }
     | { planType: InstallmentPlan.FIXED_CYCLE_AMOUNT; fixedCycleAmount: number };
 
-export interface SubscriptionsListParams extends CRUDPaginationParams {
+export type SubscriptionsListParams = CRUDPaginationParams & {
     search?: string;
     status?: SubscriptionStatus;
     mode?: ProcessingMode;
-}
+};
 
 export type ScheduledPaymentsListParams = CRUDPaginationParams;
 
-export interface SubscriptionCreateBaseParams<T extends Metadata = Metadata> {
+export type SubscriptionCreateBaseParams<T extends Metadata = Metadata> = {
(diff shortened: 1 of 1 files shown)


src/resources/Subscriptions.ts
```diff
@@ -74,19 +74,18 @@ export type SubscriptionPlanItem =
     | { planType: InstallmentPlan.FIXED_CYCLES; fixedCycles: number }
     | { planType: InstallmentPlan.FIXED_CYCLE_AMOUNT; fixedCycleAmount: number };
 
-export interface SubscriptionsListParams extends CRUDPaginationParams {
+export type SubscriptionsListParams = CRUDPaginationParams & {
```
phieronymus, line 77: Updating interfaces to types to keep things a bit more consistent across applications. For most of them there is no reason to use interfaces over types.

src/resources/Subscriptions.ts
```diff
@@ -132,26 +131,41 @@ export interface SubscriptionUpdateParams<T extends Metadata = Metadata> {
...
 
 /* Response */
-export interface SubscriptionItem<T extends Metadata = Metadata> {
+
+type PeriodParams =
+    | {
+          /**
+           * Can not be used with the cyclicalPeriod parameter
+           */
+          period: SubscriptionPeriod;
```
gdp88, line 151: should we add something like 

```
 cyclicalPeriod?: never;
 ```
 here too? Or not so needed since we'll switch to using `cyclicalPeriod ` more from now on?

gdp88 (approved):

gdp88:
just a minor comment..lgtm!

src/resources/Subscriptions.ts
phieronymus: This is not needed as the error will be triggered as it will fall in the other type when the `cyclicalPeriod` is present. Just decided that the error should be on the period rather than the cyclical period.

gdp88: alright yeah that sounds good then 👍 Resolving!

gdp88:
merging!

PR #785, opened 2023-05-15
Add total hits param

phieronymus:
Linked to https://univapay-systems.slack.com/archives/CA0U3CSQ1/p1683822852597689

### DIFF
--- a/src/resources/CRUDResource.ts
+++ b/src/resources/CRUDResource.ts
@@ -25,6 +25,10 @@ export type CRUDItemsResponse<A> = {
 };
 
 export type CRUDAOSItemsResponse<A> = CRUDItemsResponse<A> & {
+    /**
+     * Optional parameter indicating the total numbers of elements the list has.
+     * Only present when no cursor are provided when fetching the list.
+     */
     totalHits?: number;
 };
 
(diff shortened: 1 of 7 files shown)


src/resources/CRUDResource.ts
```diff
@@ -12,17 +12,22 @@ export enum CursorDirection {
...
-export interface CRUDItemsResponse<A> {
+export type CRUDItemsResponse<A> = {
     items: A[];
     hasMore: boolean;
-}
+    totalHits?: number;
+};
+
+export type CRUDAOSItemsResponse<A> = CRUDItemsResponse<A> & {
+    totalHits?: number;
```
gpapako, line 29: do we need this? Cause we just added totalHits to CRUDItemsResponse 🤔

gpapako: Actually.. maybe we should not have added it to CRUDItemsResponse? 🤔

gpapako (commented):
Just a comment. We can merge after resolving this

src/resources/CRUDResource.ts
phieronymus: Oh yeah, you are right. I added it to the wrong place and apparently forgot to remove it from there. Removing from `CRUDItemsResponse`.

gpapako (approved):
LGTM!

gpapako:
Merging!

PR #798, opened 2023-05-25
Fix subscription validation

phieronymus:
Support OR in required parameter an cyclical period for subscription creation


YusukeFukushima (approved):
Thanks a lot!

PR #804, opened 2023-05-30
Make both 'period' and 'cyclicalPeriod' optional on SubscriptionItem

gdp88:
Needed for https://github.com/univapaycast/univapay-console/issues/3287 cause otherwise typescript complains it cannot find `cyclicalPeriod` on `SubscriptionItem` when we try to use it on the merchant console side.


phieronymus (approved):
Note: This will affect the create parameters I think. But this is not so important. Let me merge that.

PR #823, opened 2023-06-20
Add cyclesLeft

YusukeFukushima:
Updates https://github.com/univapaycast/univapay-console/issues/2547
I can see `cyclesLeft` in the log, but I cannot access it on the merchant console now.
I want to use to calculate the number of cycles that alread done, so add it.

Screenshot from 2023-06-20 13-27-48


phieronymus (approved):
Right this is there 🤔 Thanks for adding it! 🙂

PR #834, opened 2023-07-04
feat(dependabot): group typescript-eslint updates

perrin4869:
Should make life easier moving forward :)


phieronymus:
Oh right! Those are always together. Thanks a lot for adding this! :meow_party:

phieronymus (approved):

PR #846, opened 2023-07-19
Add email and hidePrefilledEmail to ECFormLinkItem

YusukeFukushima:
Updates https://github.com/univapaycast/univapay-console/issues/3314
Updates https://github.com/univapaycast/univapay-form-checkout/issues/546

### DIFF
--- a/src/resources/ECFormLinks.ts
+++ b/src/resources/ECFormLinks.ts
@@ -31,7 +31,6 @@ export type ECFormLinkItem = {
     tokenType: TransactionTokenType;
     tokenOnly: boolean;
     hideCvv?: boolean;
-    email?: string;
     hidePrefilledEmail?: boolean;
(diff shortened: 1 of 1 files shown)


src/resources/ECFormLinks.ts
```diff
@@ -31,6 +31,8 @@ export type ECFormLinkItem = {
     tokenType: TransactionTokenType;
     tokenOnly: boolean;
     hideCvv?: boolean;
+    email?: string;
```
phieronymus, line 34: We do not have that. We will have to rely on the link metadata for this parameter. We only need to add `hidePrefilledEmail` 🙂

YusukeFukushima: Removed :+1: Please check again :bow:

phieronymus (approved):
Thanks!

PR #847, opened 2023-07-20
chore: support node >= 16

perrin4869:
After this we can rebase and merge #845 then make a major release
Then update the private node version :)


phieronymus (approved):
Thanks. Lets roll with that!

PR #857, opened 2023-08-02
Support jwt refresh without bearer

phieronymus:
Closes https://github.com/univapaycast/univapay-console/issues/3414


YusukeFukushima (approved):
Thanks!

PR #859, opened 2023-08-04
Update redirect for online charge

YusukeFukushima:
Updates https://github.com/univapaycast/univapay-form-checkout/issues/625


YusukeFukushima:
@phieronymus 
Just to confirm, is there no nedd to add `redirect.endpoint` to the response?
I checked the thread what is in the ticket, kingston says endpoint will be returned too.

phieronymus (approved):
Thanks!

PR #898, opened 2023-09-22
Add expirationPeriod to TransactionTokenUpdateParams

YusukeFukushima:
Updates https://github.com/univapaycast/univapay-admin-console/issues/1764

`expirationTimeShift` is also added cuz I think we'll need it in near future, but if it is not needed now, please let me know :bow:


phieronymus (approved):
Thanks. 🙂

PR #922, opened 2023-10-20
Add new alipay plus brands

phieronymus:
Linked to https://github.com/univapaycast/univapay-common/issues/1126

### DIFF
--- a/src/resources/common/enums.ts
+++ b/src/resources/common/enums.ts
@@ -142,10 +142,30 @@ export enum OnlineBrand {
     BOOST = "boost",
     RABBIT_LINE_PAY = "rabbit_line_pay",
 
+    // Alipay+ Online
+    TINABA = "tinaba",
+    NAVER_PAY = "naver_pay",
+    TOSS_PAY = "toss_pay",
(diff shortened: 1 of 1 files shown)


src/resources/common/enums.ts
```diff
@@ -142,12 +142,32 @@ export enum OnlineBrand {
...
+    KREDIVO = "kredivo",
+
     /**
      * @deprecated Use `ALIPAY_PLUS_ONLINE` instead
      */
     ALIPAY_CONNECT_ONLINE = "alipay_connect_online",
 }
 
+// Alipay+ Offline
+export enum RetailsBrands {
```
gdp88, line 160: should we go with `RetailBrands` to be consistent with this naming pattern that we use plural only for the last word?

gdp88 (approved):

gdp88:
minor comment..lgtm!

src/resources/common/enums.ts
phieronymus: Indeed, thanks for catching that! 👍

gdp88:
thanx! merging!

PR #942, opened 2023-11-14
Fix missing mapping

phieronymus:
Linked to https://univapay-systems.slack.com/archives/C0172GCPT4M/p1699940973053429

We are missing one of the HTTP mapping for charge errors.


gpapako (approved):
👍

PR #954, opened 2023-11-30
Fix alipay+ brand keys

phieronymus:
Seems that most of the keys are used were wrong. Asked API and got some magical answers


gdp88 (approved):

gdp88:
alright hopefully they are correct at least now! merging!

PR #958, opened 2023-12-06
Add refer routes

phieronymus:
Linked to https://github.com/univapaycast/gyron-payments-api/issues/6606

Updates https://github.com/univapaycast/univapay-form-checkout/issues/722

### DIFF
--- a/src/resources/Subscriptions.ts
+++ b/src/resources/Subscriptions.ts
@@ -62,6 +62,7 @@ export interface ScheduledPaymentItem {
     isLastPayment: boolean;
     createdOn: string;
     successfulPaymentDate?: string;
+    terminateWithStatus?: SubscriptionStatus;
 }
 
 export type SchedulePaymentListItem = ScheduledPaymentItem;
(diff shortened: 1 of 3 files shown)


src/resources/Subscriptions.ts
```diff
@@ -339,6 +340,20 @@ export class Subscriptions extends CRUDResource {
         return this._list(data, auth, { storeId });
     }
 
+    private _suspend?: DefinedRoute;
+    suspend(storeId: string, id: string, data?: SendData<void>, auth?: AuthParams): Promise<ResponsePayment> {
```
gpapako, line 344: I think for suspend / unsuspend we also need the merchantId? 🤔

gpapako (approved):
LGTM! Just a comment in case we need to add something there

src/resources/Subscriptions.ts
phieronymus: Merchant id is for the private SDK. No need for the public one.

PR #959, opened 2023-12-07
Fix customer service route name

phieronymus:
I used the internal route names instead of the public one 🙇

### DIFF
--- a/src/resources/Refers.ts
+++ b/src/resources/Refers.ts
@@ -35,7 +35,7 @@ export type ParsedAuthenticateToken = {
 };
 
 export class Refers extends CRUDResource {
-    static routeBase = "/stores/:storeId/refers";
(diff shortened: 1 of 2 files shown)


src/resources/Refers.ts
```diff
@@ -35,7 +35,7 @@ export type ParsedAuthenticateToken = {
 };
 
 export class Refers extends CRUDResource {
-    static routeBase = "/stores/:storeId/refers";
+    static routeBase = "(/stores/:storeId/)customers";
```
gpapako, line 38: `(/stores/:storeId)/customers`?

gpapako (approved):
Just one small comment

src/resources/Refers.ts
phieronymus: Thanks!

PR #964, opened 2023-12-13
Fix types used in checkouts

phieronymus:
Fixes the types according to the checkout TODOs to prevent unwanted casts


gdp88 (approved):

gdp88:
lgtm! merging!

PR #966, opened 2023-12-18
Add customerReferLinkEnabled parameter

phieronymus:
This parameter controls the display of one of the template parameter. We will need it for tomorrow release.

Still have some question to ask to Camille and Kingston about it so opening it as the draft


gpapako (approved):
👍

PR #971, opened 2023-12-21
Add sbscription statuses filter

phieronymus:
Linked to https://univapay-systems.slack.com/archives/CA0U3CSQ1/p1703159104669109

### DIFF
--- a/src/resources/Subscriptions.ts
+++ b/src/resources/Subscriptions.ts
@@ -78,6 +78,7 @@ export type SubscriptionPlanItem =
 export type SubscriptionsListParams = CRUDPaginationParams & {
     search?: string;
     status?: SubscriptionStatus;
+    statuses?: SubscriptionStatus[];
     mode?: ProcessingMode;
 };
 


src/resources/Subscriptions.ts
```diff
@@ -78,6 +78,7 @@ export type SubscriptionPlanItem =
 export type SubscriptionsListParams = CRUDPaginationParams & {
     search?: string;
     status?: SubscriptionStatus;
```
gpapako, line 80: Do we still keep the `status` ? Or are we gonna switch to using `statuses` only from now on?

gpapako (commented):
Just a question and then we can merge

src/resources/Subscriptions.ts
phieronymus: API did not remove it so we have no reason to not support it any longer if a customer wants to use it.

gpapako: yes, of course. Was just checking if the API removed it or not. 👍

gpapako (approved):
👍

PR #1002, opened 2024-02-06
fix(coveralls): finish parallel jobs explicitly to avoid flakeyness

perrin4869:


perrin4869:
I feel like I may be missing something: https://docs.coveralls.io/parallel-builds
not sure what that `if: $` is supposed to be there and I get the feeling that it somehow got interpolated wrong when it's rendering the page lol
but this seems to fix the issue of the flakeyness of coveralls on the CI matrix

gdp88 (approved):

gdp88:
sure let's go with that! lgtm!

PR #1005, opened 2024-02-08
Fix renew event

phieronymus:
Updates https://github.com/univapaycast/univapay-console/issues/3580

Seems that the webhook event was the wrong one after discussing with API. I got confused with the payment event.


gdp88 (approved):

gdp88:
lgtm! merging!

PR #1075, opened 2024-06-04
Add three ds fields in platform user defaults card configuration

gpapako:
Related to: https://github.com/univapaycast/univapay-admin-console/issues/1887


phieronymus:
Oh thnaks! Seems that I forgot it there 🙇

phieronymus (approved):

PR #1076, opened 2024-06-04
Rename 'subscriptions' and 'charge' fields on batch subscription resp…

gdp88:
Change that follows from here https://github.com/univapaycast/univapay-form-checkout/pull/929#discussion_r1623991226

Needs https://github.com/univapaycast/services-typescript/pull/3073 to be merged first.


phieronymus (approved):
Seems that the backend is merged as well 🙂

PR #1077, opened 2024-06-10
Add retryInterval to ScheduledPaymentItem

gpapako:
Related to: https://github.com/univapaycast/univapay-admin-console/issues/1902


gpapako:
@phieronymus Please let me know if this is not the right way to use the field. 🙇

phieronymus:
@gpapako This is fine here. Just be sure it is also on the private node in the update parameters as well :meow_coffee:

phieronymus (approved):

PR #1082, opened 2024-06-11
Add updatedOn to refund item

gdp88:
Needed for https://github.com/univapaycast/univapay-admin-console/issues/1898

### DIFF
--- a/src/resources/Refunds.ts
+++ b/src/resources/Refunds.ts
@@ -63,6 +63,7 @@ export interface RefundItem<T extends Metadata = Metadata> {
     metadata?: T;
     mode: ProcessingMode;
     createdOn: string;
+    updatedOn: string;
 }
 
 export type RefundListItem = RefundItem;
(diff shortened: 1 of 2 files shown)


src/resources/Refunds.ts
```diff
@@ -63,6 +63,7 @@ export interface RefundItem<T extends Metadata = Metadata> {
     metadata?: T;
     mode: ProcessingMode;
     createdOn: string;
+    updatedOn: string;
```
phieronymus, line 66: You may want to update the private SDK instead (`RefundAdminItem`) as this is an admin-only field.

gdp88: oh alright in that case I guess no changes needed here or the private sdk since it seems like `RefundAdminItem` can have the `updatedOn` field already:

```
export interface RefundAdminItem extends RefundItem, WithUpdatedOn {
    merchantId: string;
}
```

gdp88:
let me close this PR then cause I guess we don't need to make any changes here after all. Thanx for checking!

PR #1185, opened 2024-10-21
Modify API to accommodate cookie authentication

gpapako:
Related to: https://github.com/univapaycast/call-center/issues/12

I decided to modify the API to accommodate for the case of cookie authorization as per the call-center new API documentation.

https://github.com/univapaycast/call-center/blob/develop/openapi/openapi.yaml

### DIFF
--- a/src/api/RestAPI.ts
+++ b/src/api/RestAPI.ts
@@ -117,6 +117,7 @@ export interface AuthParams {
     secret?: string;
     idempotentKey?: string;
     origin?: string;
+    useCredentials?: boolean;
 
     // Deprecated
     authToken?: string;
@@ -244,6 +245,7 @@ export class RestAPI extends EventEmitter {
         const params: RequestInit = {
             headers: this.getHeaders(data, auth, payload, acceptType),
             method,
+            ...(auth?.useCredentials ? { credentials: "include" } : {}),
         };
 
         const request: Request = new Request(
--- a/test/specs/api.specs.ts
+++ b/test/specs/api.specs.ts
@@ -299,6 +299,45 @@ describe("API", function () {
         }
     });
 
+    it("should set request credentials to include when userCredentials is passed through AuthParams", async function () {
+        if (process.version < "v18.00") {
+            // Request credentials was introduced in v18.00, so skip older versions
+            this.skip();
+        }
+        const loginMock = fetchMock.get(`${testEndpoint}/login`, {
+            status: 200,
+            body: okResponse,
+            headers: {
+                "Content-Type": "application/json",
+                "Access-Control-Allow-Credentials": "true",
+                "Set-Cookie": `SESSIONID=12345; Path=/; SameSite=none`,
+            },
+        });
+
+        const api: RestAPI = new RestAPI({ endpoint: testEndpoint });
+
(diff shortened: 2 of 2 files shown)


gpapako:
Converting to draft. My understanding of how cookies work was probably off. 
Set-Cookie header should be setting a cookie on the browser and then that cookie would be automatically returned on every request as long as the path matches what is described in the cookie. 

Might not need this PR at all. Let me investigate a bit more.

src/api/RestAPI.ts
```diff
@@ -244,6 +244,7 @@ export class RestAPI extends EventEmitter {
         const params: RequestInit = {
             headers: this.getHeaders(data, auth, payload, acceptType),
             method,
+            credentials: "include",
```
gpapako, line 247: this seems to be the only change we need on the frontend

test/specs/api.specs.ts
```diff
@@ -299,6 +299,40 @@ describe("API", function () {
         }
     });
 
+    // TODO: figure out how to test that a cookie is automatically set by `set-cookie` header
```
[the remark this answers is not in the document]
phieronymus: Are there some cookie mock libraries?

src/api/RestAPI.ts
phieronymus: We should probably use the `auth` parameter to set this 🙂 Testing all the cases may take a while otherwise.

phieronymus (commented):
Looking good. Two minor comments

src/api/RestAPI.ts
```diff
@@ -244,6 +245,7 @@ export class RestAPI extends EventEmitter {
         const params: RequestInit = {
             headers: this.getHeaders(data, auth, payload, acceptType),
             method,
+            ...(auth?.useCredentials ? { credentials: "include" } : {}),
```
gpapako, line 248: made the `credentials`  request parameter optional so that it can be used only for the API that uses cookie authentication.

gpapako:
I can't seem to be able to create a test with what we want here. 
Basically, `Set-Cookie` should store that cookie (or cookies) on the browser.
And then when we call the same domain, the request should automatically contain the `Cookie` header with said cookie.
I added something more that seems to be important which is the `credentials: true` parameter in each request.

The issue here seems to be that storing the cookie and setting the `Cookie` header on requests is a browser functionality that I cannot somehow mock. I can `manually` add a header in the request while testing but this is not the same thing. 

Also the cookie is supposed to be `http-only` which should not be accessible to Typescript / Javascript.

phieronymus (approved):
Perfect! Thanks a lot!

phieronymus:
Somehow the new test is failing on node 16 🤔

gpapako:
Let me try to investigate this a bit

gpapako:
It seems that there is a lack of support on node 16 for the `credentials` parameter. 
https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials
Supported from node 18 and higher? 🤔

phieronymus:
I was looking into it and it seems that most browser will support it so no real problem.

Dropping node 16 is not the plan for this PR so would it be possible to skip this test for node 16 with a comment? We will fix this once we drop 16 for 22.

PR #1223, opened 2024-12-24
Add variant to enum

gpapako:
Related to: https://github.com/univapaycast/univapay-form-checkout/pull/1115

### DIFF
--- a/src/resources/Charges.ts
+++ b/src/resources/Charges.ts
@@ -43,6 +43,7 @@ export enum ChargeThreeDsMode {
     SKIP = "skip",
     IF_AVAILABLE = "if_available",
     REQUIRE = "require",
+    PROVIDED = "provided",
 }
 
 export type ChargeThreeDsIssuerToken = {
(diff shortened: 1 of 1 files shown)


src/resources/Charges.ts
```diff
@@ -42,6 +42,7 @@ export enum ChargeThreeDsMode {
     FORCE = "force",
     SKIP = "skip",
     IF_AVAILABLE = "if_available",
+    REQUIRE = "require",
```
kgston, line 45: Technically speaking, there is 1 more called `provided`, but it is a read only value, meaning to say, you can never set that value in the request. You can only receive as a response. You might want to add it if parsing a non existing enum value will cause your code to crash.

kgston (approved):

src/resources/Charges.ts
gpapako: sure! sounds good! Better have it there to be on the safe side

kgston (approved):

gpapako:
Merging

PR #1225, opened 2025-01-14
Add types for call center store

gdp88:
Moving the store enums/types here to mitigate circular references on the call center side.


phieronymus:
The call center and gpp store are two different resources. Best would be to create a new file to prevent confusion.

phieronymus:
Also. This needs to be added to the private ask, not the public one as this is a private api.

gdp88:
Closing cause the types need to be added to the private sdk (PR: https://github.com/univapaycast/univapay-node-private/pull/1445)

PR #1228, opened 2025-02-07
Allow body transfer encoding overrides

phieronymus:
It was asked for the services multiple time and backend was changed but it is only a matter of time before customer want to do the same as the web specifications allow for pretty wild usage.

### DIFF
--- a/src/api/RestAPI.ts
+++ b/src/api/RestAPI.ts
@@ -163,7 +163,7 @@ export type ApiSendOptions = {
     /**
      * Overrides default body transfer encoding of the request.
      *
-     * Note: Can not for for GET and HEAD request
+     * Note: Can not for GET and HEAD request
      *
      * Default is "message-body" for DELETE, HEAD and GET and "entity-body" for all other call methods.
      */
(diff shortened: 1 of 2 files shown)


src/api/RestAPI.ts
```diff
@@ -247,9 +258,19 @@ export class RestAPI extends EventEmitter {
...
+        const {
+            acceptType,
+            keyFormatter = toSnakeCase,
+            ignoreKeysFormatting = ["metadata"],
+            bodyTransferEncoding,
+        } = options;
+
+        const payload: boolean =
+            bodyTransferEncoding === "entity"
+                ? true && ![HTTPMethod.GET, HTTPMethod.HEAD].includes(method)
```
phieronymus, line 270: As per specification GET and HEAD can never have a payload.

YusukeFukushima (approved):
Thanks :bow: 
I'll not use it at this time, but whenever I may use this

phieronymus:
@YusukeFukushima Yeah. It is best not to use a body for the delete call method as it is not recommended. This if for customer who do not have the choice given their API.

src/api/RestAPI.ts
gdp88: Do we need the `true &&` part here? 🤔

gdp88 (approved):

gdp88:
minor comments..lgtm!

src/api/RestAPI.ts
```diff
@@ -157,6 +159,15 @@ export type ApiSendOptions = {
      * List of key to ignore formatting on when parsing the response
      */
     ignoreKeysFormatting?: string[];
+
+    /**
+     * Overrides default body transfer encoding of the request.
+     *
+     * Note: Can not for for GET and HEAD request
```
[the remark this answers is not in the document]
phieronymus: Ups. Thanks.

phieronymus:
Indeed, the true was not necessary, thanks for catching this 🙂

gdp88:
thanx! merging!

PR #1229, opened 2025-04-01
Add Aeon pay gateway

gpapako:
Related to: https://github.com/univapaycast/univapay-admin-console/issues/2052

### DIFF
--- a/src/resources/common/enums.ts
+++ b/src/resources/common/enums.ts
@@ -61,6 +61,7 @@ export enum ProcessingMode {
 }
 
 export enum QRBrand {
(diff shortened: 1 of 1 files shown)


src/resources/common/enums.ts
```diff
@@ -104,6 +105,7 @@ export enum QRBrand {
 }
 
 export enum QRGateway {
+    AEON_PAY = "aeon_pay",
```
gpapako, line 108: I guess this is needed too 🤔

phieronymus (approved):

PR #1230, opened 2025-05-16
Add 3DS token/charge errors

gdp88:
Required for https://github.com/univapaycast/univapay-form-checkout/issues/1154


src/errors/parser.ts
```diff
@@ -8,35 +8,36 @@ import { PathParameterError } from "./PathParameterError.js";
...
         302: ResponseErrorCode.CardExpirationMonthInvalid,
         303: ResponseErrorCode.CardExpirationYearInvalid,
-        304: ResponseErrorCode.CardExpired,
+        304: isThreeDsError ? ResponseErrorCode.ThreeDSCardExpired : ResponseErrorCode.CardExpired,
         305: ResponseErrorCode.CardCVVInvalid,
         306: ResponseErrorCode.CardRejected,
-        307: ResponseErrorCode.CardInvalid,
+        307: isThreeDsError ? ResponseErrorCode.ThreeDSCardInvalid : ResponseErrorCode.CardInvalid,
         308: ResponseErrorCode.ChargeInvalidData,
-        309: ResponseErrorCode.ProcessingError,
```
gdp88, line 21: remove `309` mapping since those errors are now handled in common

phieronymus: We can do the mapping for the messages here though, so that we have a consistent error message in case we need to show it several times.

gdp88:
This should be ready now based on our new interpretation of the specs. Removing `Draft`.

src/errors/APIError.ts
```diff
@@ -195,13 +195,17 @@ export enum ResponseErrorCode {
...
+    DisputedTransaction = "DISPUTED_TRANSACTION",
     MissingThreeDSServerTransID = "MISSING_THREE_DS_SERVER_TRANS_ID",
     PaymentTypeNotSupported = "PAYMENT_TYPE_NOT_SUPPORTED",
     InvalidThreeDsStatus = "INVALID_THREE_DS_STATUS",
+    CardNotEnrolledInThreeDS = "CARD_NOT_ENROLLED_IN_THREE_DS",
     ThreeDsCouldNotBeCompleted = "THREE_DS_COULD_NOT_BE_COMPLETED",
     InternalError = "INTERNAL_ERROR",
 
+    // Charge 3DS default error
+    ThreeDsDefaultError = "THREE_DS_DEFAULT_ERROR",
```
phieronymus, line 207: Can we name it `ThreeDsProcessingError` or `GenericThreeDsProcessingError`. Maybe I used the `default` word before but this is not a default.

src/errors/parser.ts
```diff
@@ -8,35 +8,35 @@ import { PathParameterError } from "./PathParameterError.js";
 import { RequestParameterError } from "./RequestParameterError.js";
 import { RequestError, RequestResponseBaseError, ResponseError } from "./RequestResponseError.js";
 
-export const getProcessingErrorCodeByStatus = (status: number): string => {
+export const getProcessingErrorCodeByStatus = (status: number, isThreeDsError?: boolean, message?: string): string => {
```
phieronymus, line 11: Very nitpicky on my part but would it be possible to have `message` before `isThreeDsError`?

Also, feel free to ignore this but adding an `options?: { isThreeDsError?: boolean }` feels more future-proof and would also document the code better than just a boolean.

src/errors/parser.ts
```diff
@@ -8,35 +8,35 @@ import { PathParameterError } from "./PathParameterError.js";
...
         303: ResponseErrorCode.CardExpirationYearInvalid,
-        304: ResponseErrorCode.CardExpired,
+        304: isThreeDsError ? ResponseErrorCode.ThreeDsDefaultError : ResponseErrorCode.CardExpired,
         305: ResponseErrorCode.CardCVVInvalid,
         306: ResponseErrorCode.CardRejected,
-        307: ResponseErrorCode.CardInvalid,
+        307: isThreeDsError ? ResponseErrorCode.ThreeDsDefaultError : ResponseErrorCode.CardInvalid,
         308: ResponseErrorCode.ChargeInvalidData,
-        309: ResponseErrorCode.ProcessingError,
+        309: isThreeDsError ? ResponseErrorCode.ThreeDsDefaultError : ResponseErrorCode[message],
```
phieronymus, line 21: Please display the processing error when the message is not found in the error code to keep the current handling.

Also, please add the missing messages in the `ResponseErrorCode` if any 🙇

gdp88:
@phieronymus tried to address ur comments..let me know if something else needs to be changed too 🙇

phieronymus:
Looking great, thanks!

phieronymus (approved):

PR #1231, opened 2025-05-27
Kepp existing translation even if 3DS error

gdp88:
Based on latest feedback here https://univapay-systems.slack.com/archives/C03CECHJDSQ/p1747383735664759


phieronymus (approved):
Thanks!

PR #1234, opened 2025-06-25
Update enums for direct debit

gpapako:
Partially addresses: https://github.com/univapaycast/univapay-console/issues/3897

### DIFF
--- a/src/resources/BankAccounts.ts
+++ b/src/resources/BankAccounts.ts
@@ -15,7 +15,7 @@ export enum BankAccountStatus {
 }
 
 export enum BankAccountType {
-    REGULAR = "regular",
+    CHECKING = "checking",
     SAVINGS = "savings",
 }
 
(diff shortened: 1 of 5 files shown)


gpapako:
https://github.com/univapaycast/univapay-direct-debit/blob/master/specs/api.yml#L1764

Screenshot 2025-06-25 at 11 57 23
Screenshot 2025-06-25 at 11 57 51

src/resources/BankAccounts.ts
```diff
@@ -15,7 +15,7 @@ export enum BankAccountStatus {
 }
 
 export enum BankAccountType {
-    REGULAR = "regular",
+    CHECKING = "checking",
     SAVINGS = "savings",
 }
```
phieronymus, line 20: We should likely not update this. Relying on the direct debit status would be best as this will affect unrelated part of the application.

gpapako: Maybe it's confusing through the commits, but it is not updated after all. Only the values for `DirectDebitTransferDateType` are updated

gpapako: In my latest commit, I reverted the changes I had performed in the first commit. Sorry for the confusion 🙇

phieronymus: Ups, my bad. Maybe I was looking at the wrong thing. My brain is fairly slow this week.

PR #1244, opened 2025-12-17
Update CI

phieronymus:
Updating CI to match the new npm expectations.


rporter-upc (approved):
Not seeing any obvious problems here and for peace of mind, actions all appear to be running fine. Only one not tested is publish, but looking closely at the configs, again, no obvious problems. Hopefully it runs smoothly 🤞

phieronymus:
Yes, publish is onntag creation. I'll be on that today.

